PR #2, opened 2022-04-22
Log errors in non-dereffed futures

mainej:
Exceptions in futures that are never dereffed are essentially lost, unless they are explicitly logged. They are lost because Clojure catches them and holds them, waiting to re-throw them when the future is dereffed. This adds logging for a few futures in lsp4clj.core.

It also deprecates some buggy helper functions. These functions are intended for debugging the input to and output from LSP. However, they have several deficiencies.

First, LSP has built-in tools for tracing its I/O. Prefer the 6 argument builder for an LSP Launcher, which accepts a PrintWriter for tracing.
E.g.

```clojure
(let [tracer (java.io.PrintWriter. (io/writer (io/file "path/to/a/log/file")))]
 (Launcher/createLauncher server ClojureLanguageClient is os false tracer))
```

Second, they will throw a "write end dead" java.io.IOException. It doesn't happen on my MacBook Pro, but happens reliably on Github Actions (ubuntu-latest) after about 30 seconds of use.

### DIFF
--- a/server/src/lsp4clj/core.clj
+++ b/server/src/lsp4clj/core.clj
@@ -102,7 +102,11 @@
 
   (^void didSave [_ ^DidSaveTextDocumentParams params]
     (future
-      (handle-notification params feature-handler/did-save handler))
+      (try
+        (handle-notification params feature-handler/did-save handler)
+        (catch Throwable e
+          (logger/error e)
+          (throw e))))
     (CompletableFuture/completedFuture 0))
 
   (^void didClose [_ ^DidCloseTextDocumentParams params]
@@ -212,7 +216,11 @@
   WorkspaceService
   (^CompletableFuture executeCommand [_ ^ExecuteCommandParams params]
     (future
-      (handle-notification params feature-handler/execute-command handler))
+      (try
+        (handle-notification params feature-handler/execute-command handler)
+        (catch Throwable e
+          (logger/error e)
+          (throw e))))
     (CompletableFuture/completedFuture 0))
 
   (^void didChangeConfiguration [_ ^DidChangeConfigurationParams params]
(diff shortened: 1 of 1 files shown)


Cyrik:
Looks good 👍

server/src/lsp4clj/core.clj
```diff
@@ -102,7 +102,11 @@
 
   (^void didSave [_ ^DidSaveTextDocumentParams params]
     (future
-      (handle-notification params feature-handler/did-save handler))
+      (try
+        (handle-notification params feature-handler/did-save handler)
+        (catch Throwable e
+          (logger/error e)
+          (throw e))))
```
ericdallo, line 109: Usually, I don't like to log and throw, most of the time you want to do one or other, in this case, throwing would already make lsp4j log, right?

ericdallo: Hum, not sure because the way we custom log, probably something we could fix at a higher level of lsp4clj, seems ok to keep this way for now

ericdallo (approved):

server/src/lsp4clj/core.clj
mainej: Unfortunately, no. Usually you're right, when a thread has an uncaught exception it's automatically logged. But `future`s are special. They catch all exceptions and hold them, waiting to re-throw them when the future is dereffed (which we never do for this future.) So, since there aren't any uncaught exceptions, there's no automatic logging. I think (though I'm not sure) that if we changed `future` to `async/go` or `async/thread`, exceptions would be thrown.

You're right though, since this future is thrown away whether its body completes successfully or not, there's no real benefit to re-throwing. It'd be sufficient to log and return nil.

ericdallo: Yeah, forgot this was a future, even so, as you said, logging in a future is enough indeed

PR #3, opened 2022-06-21
Debean client capabilities completion item resolve support

mainej:
To support https://github.com/clojure-lsp/clojure-lsp/issues/1068

### DIFF
--- a/server/src/lsp4clj/coercer.clj
+++ b/server/src/lsp4clj/coercer.clj
@@ -51,7 +51,6 @@
      CreateFile
      RenameFile
      RenameOptions
-     ResponseErrorCode
      PrepareRenameResult
      SaveOptions
      SemanticTokens
@@ -81,7 +80,8 @@
      ResponseErrorException)
    (org.eclipse.lsp4j.jsonrpc.messages
      Either
-     ResponseError)))
+     ResponseError
+     ResponseErrorCode)))
 
 (set! *warn-on-reflection* true)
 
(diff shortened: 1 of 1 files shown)


server/src/lsp4clj/coercer.clj
```diff
@@ -638,6 +637,11 @@
 (s/def :client-capabilities/signature-help (s/and ::debean
                                                   (s/keys :opt-un [:client-capabilities/signature-information])))
 
+(s/def :client-capabilities/resolve-support ::debean)
+
+(s/def :client-capabilities/completion-item (s/and ::legacy-debean
```
ericdallo: Since we are touching this, can't we use `::debean` instead of `::legacy-debean`?

ericdallo: The idea is to replace ::legacy-debean with ::deabean but changing all places would be hard to test, so I like to change when we need to fix a bug or something like this one and test that feature only

ericdallo (approved):

server/src/lsp4clj/coercer.clj
mainej: I can make that switch... but why do we still have `::legacy-debean`? And why do we use `j/from-java`, instead of `j/from-java-deep`, especially for data objects like client capabilities?

ericdallo: We had some issues with the older debean, so I deprecated and created the new `::debean`,but I didn't change all places to use `::debean` to avoid lots of possible issues.

> And why do we use j/from-java, instead of j/from-java-deep

I don't know, didn't know that exists

mainej: I switched from `::legacy-debean` to `::debean` and it's working fine, so this is ready for a final review. I played with `from-java-deep` a little. It's probably the right tool to use, but I couldn't get it to work... it was having some problems with reflection on JsonObject or JsonElement. But it's probably worth getting working—it'd simplify a lot of input coercion.

ericdallo (approved):

PR #10, opened 2022-07-06
notebook support

Cyrik:
I had to put a snapshot version of lsp4j into the repo temporarily since it's not published yet. I'll remove that once they publish.
Otherwise, this seems to do the trick 🥳


server/src/lsp4clj/coercer.clj
```diff
@@ -616,6 +619,7 @@
 (s/def :client-capabilities/document-link ::legacy-debean)
 (s/def :client-capabilities/formatting ::legacy-debean)
 (s/def :client-capabilities/implementation ::legacy-debean)
+(s/def :client-capabilities/notebook-document ::legacy-debean)
```
ericdallo: Since it's a new code, can't we use `::debean` instead of the legacy one?

Cyrik: Sure, I'll have a look at that. Just got lazy here 😅

mainej:
@Cyrik I think part or most of this is obsolete after #13. Can it be closed? Or do you want to update it? If you want to chat about how to implement `receive-notification` for the `"notebookDocument/did*"`  methods, let me know. Do you need any of the coercion? It looks like that was just to generate Java objects.

Cyrik:
We're currently depending on this for our LSP, but the plan is to move to the new version once I find some time.
It's enough that I still have the branch though so the PR can go.

PR #11, opened 2022-07-06
Discard stdout, instead of sending it to clients

### LINKED ISSUE #9: Capture and log stdout
The cause of https://github.com/clojure-lsp/clojure-lsp/issues/1080 was that clojure-lsp was accidentally writing to stdout. This was found in https://github.com/clojure-lsp/clojure-lsp/pull/1083 and fixed in https://github.com/clojure-lsp/clojure-lsp/pull/1087.

@ericdallo suggested the following changes in lsp4clj to prevent similar problems in the future:

1. lsp4clj should wrap all of its calls into clojure-lsp and other servers with `with-out-str`. This is because cloure-lsp uses `System.out` to communicate with clients (and I believe all other servers do too). If the server accidentally prints to stdout while it's processing a request, `with-out-str` will avoid letting that output leak into the server/client communication channel.
2. lsp4clj should log any non-blank output strings.

mainej:
Language servers implemented with lsp4clj tend to send their output over
stdout. Clients read this output and parse it as JSON-RPC. So, it
shouldn't contain anything besides what the server intended to send.

But, if any server prints to stdout, even accidentally, that output will
be intermingled with the JSON-RPC messages.
https://github.com/clojure-lsp/clojure-lsp/issues/1080 was one such
issue. Interestingly Calva users experienced this issue, but Emacs users
didn't, suggesting that some clients handle unexpected data in the
server output better than others.

This patch adds a helper that captures stdout and ~logs~ discards it, instead of
sending it to clients. Fixes #9.

It protects all of lsp4clj's uses of ILSPFeatureHandler and
ILSPProducer. This includes:
- Uses of the feature handler and producer in LSPServer.
- Uses of the feature handler and producer in LSPTextDocumentService.
- Uses of the feature handler and producer in LSPWorkspaceService.

It does not protect any of the following:
- Server code that runs before `initialize`, or any async processes
  started during this time. (e.g. cloure-lsp.server/run-server! and
  spawn-async-tasks!)
- Any code that wraps ILSPFeatureHandler or ILSPProducer (e.g.
  ClojureLSPFeatureHandler and ClojureLspProducer)
- Any extensions of the Server or Services (e.g. ClojureLanguageServer)
- Any code that uses the wrappers or extensions (e.g. ClojureLspServer)

Servers should make efforts to protect against accidental output to
stdout in all of these scenarios, by using the helper or a similar one.

~There are things to be aware of with the new helper.~

~It generates a new StringWriter and logs its contents when the body
returns. Any async processes started in the body will use the
StringWriter for their stdout, protecting the clients.  But, beware that
any output the async processes generate (after their invoker returns)
will not be logged.~

~To ensure that these processes generate logs, they should all call the
helper (or a similar one) *within* their body (i.e. inside `future`,
`go`, `thread`, etc.). This will generate a separate StringWriter, which
will be logged and discarded when the body finishes. Failing to do this
is more serious than simply not seeing a few logs; it can lead to memory
leaks. If the async process runs forever (e.g. `go-loop` or `(future 
(while true )`) it will hold a reference to the StringWriter generated
by the outer helper forever. If it starts filling it with output, that
output will never be garbage collected (and never be logged).~

~Because of this danger, in the future the helper will probably be
changed to discard output (see
https://stackoverflow.com/questions/55628999/java-processbuilder-how-to-suppress-output-instead-of-redirecting-it)
rather than logging it. But, we're starting with logging so that clients
can audit their output.~

### DIFF
--- a/server/src/lsp4clj/core.clj
+++ b/server/src/lsp4clj/core.clj
@@ -76,7 +76,7 @@
         ([^bytes b])
         ([^bytes b, off, len])))))
 
-(defmacro capturing-stdout
+(defmacro discarding-stdout
   "Evaluates body in a context in which writes to *out* are discarded."
   [& body]
   `(binding [*out* null-output-stream-writer]
@@ -111,16 +111,16 @@
          [^ILSPFeatureHandler handler]
   TextDocumentService
   (^void didOpen [_ ^DidOpenTextDocumentParams params]
-    (capturing-stdout
+    (discarding-stdout
       (handle-notification params feature-handler/did-open handler)))
 
   (^void didChange [_ ^DidChangeTextDocumentParams params]
-    (capturing-stdout
+    (discarding-stdout
       (handle-notification params feature-handler/did-change handler)))
 
   (^void didSave [_ ^DidSaveTextDocumentParams params]
     (future
-      (capturing-stdout
+      (discarding-stdout
         (try
           (handle-notification params feature-handler/did-save handler)
           (catch Throwable e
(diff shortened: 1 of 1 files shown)


mainej:
An alternative solution would be to make this helper discard stdout. Then servers could use their own version, which either discarded or logged, as they felt appropriate.

Still, there's a part of me that wants to use the logging version, even just for a little while. I want to see if any stdout is reported in the integration tests.

mainej:
Originally this PR logged stdout, but in the process of writing it I realized that introduced the potential for memory leaks. To avoid that, it has been re-written to discard stdout.

mainej:
(I'm not interested in seeing stdout in the integration tests anymore because I think the flakiness is [solved](https://github.com/clojure-lsp/clojure-lsp/pull/1109).)

server/src/lsp4clj/core.clj
```diff
@@ -69,6 +69,19 @@
 
 (set! *warn-on-reflection* true)
 
+(def null-output-stream-writer
+  (java.io.OutputStreamWriter.
+    (proxy [java.io.OutputStream] []
+      (write
+        ([^bytes b])
+        ([^bytes b, off, len])))))
```
ericdallo, line 77: Usually when using GraalVM we avoid adding top-level defs as it makes graal build slower, but this is a simple class instantiation so I guess it's ok.

server/src/lsp4clj/core.clj
```diff
@@ -69,6 +69,19 @@
...
+
+(defmacro capturing-stdout
```
ericdallo: Maybe
```suggestion
(defmacro discarding-stdout
```

ericdallo (approved):

server/src/lsp4clj/core.clj
mainej: Hmm... I didn't know that. Do you know why that slows it down? Anyway, I'll leave this as is unless you say otherwise.

server/src/lsp4clj/core.clj
mainej: I like it!

Issue #1, opened 2022-04-14
Support for more communication channels


SevereOverfl0w:
- ~~stdio: uses stdio as the communication channel.~~
-  pipe: use pipes (Windows) or socket files (Linux, Mac) as the communication channel. The pipe / socket file name is passed as the next arg or with --pipe=.
-  **socket: uses a socket as the communication channel. The port is passed as next arg or with --port=.**
-  node-ipc: use node IPC communication between the client and the server. This is only support if both client and server run under node.

I'm particularly interested in socket because it allows starting a server in a running JVM to expose runtime information that static analysis might not catch.

PR #12, opened 2022-07-14
Remove lsp4j on v1 branch

### LINKED ISSUE #8: Replace lsp4j with internal communication logic
Discussing with @mainej, we realized that lsp4j is not helping that much lsp4clj or clojure-lsp.

The main reason we use it IMO, it's because lsp4j implements the json parsing and stdout/stderr LSP jsonrpc communication layer, making the client just need to implement an interface handling the request methods and notifications easily.
But lsp4j has some drawbacks like:

- it uses a lot of reflection on the backend, making it harder to work with Graalvm, since Graal requires to tell what classes exist and what it's used.
- Unnecessary dependencies like xtend, guava, gson, and others.
- we already have a coercion logic on our side, so having the POJO classes is not needed actually if we implement the communication in Clojure
- We can't customize how async is handled, lsp4j just gives us the CompletableFuture for each request method being handled, we can't customize to use clojure.core.async or anything custom
- It's easy to cause a bug on the communication layer, as we could see on https://github.com/clojure-lsp/clojure-lsp/issues/1080 after we accidentally printed to stdout and the server stopped working.

I think we could think about how we can achieve the same lsp4j offers to us using only Clojure

mainej:
Replaces lsp4j with a Clojure implementation. The full list of dependencies is now clojure, core.async, cheshire, and camel-snake-kebab.

Also merges the lsp4clj-server and lsp4clj-protocol projects into one top-level project, lsp4clj.

clojure-lsp has an [accompanying PR](https://github.com/clojure-lsp/clojure-lsp/pull/1117).

@Cyrik, as stated in the CHANGELOG, other users (you, I think) of lsp4clj are encouraged to upgrade. You can follow that clojure-lsp PR as guidance on the changes that'll need to be made. They're fairly extensive, but I believe will make lsp4clj a more stable platform for LSP development.

Notable changes:

- As described in #8, servers can choose which requests and notifications they respond to by implementing the `lsp4clj.server/receive-request` and `lsp4clj.server/receive-notification` multimethods. (In contrast to the protocols based approach we had before.) This gives servers much more flexibility. They can implement new parts of the spec at their leisure, or according to the needs of their language. They can also choose when and how to implement asynchronicity. The tradeoff is that much of the `spec` checking and logging moves back to clojure-lsp and other servers. The servers also have to know and use the names of the LSP requests/notifications. Personally, I find `clojure-lsp.server` code to be more readable now, even though it has a few more responsibilities.
- Input maps are all coerced to kebab-case. The server receives whatever the client sent, which should follow the LSP spec. 
- Output specs still exist, but do only very lightly conformation. Of what conformation remains, most of it converts keywords to enum values. There are a few places where more complex conformation happens. Honestly, I'd like to remove almost all conformation. I think it will be easier if you can look up what to output based on the LSP spec, not the LSP spec, plus a bunch of reasoning about coercion.
- It is much less code to initialize a server. None of it is Java interop.
- We have a path to working on #1. We need code that converts a socket into a pair of input and output channels. Otherwise, much of the code can stay the same.

Fixes #8.


mainej:
And @Cyrik—I meant to mention... when you're upgrading if you'd like a little direction, I'd be happy to chat. Perhaps we can turn our discussion into a Usage section in the README.

mainej:
Sorry for the noise, but I forgot our plan for how to deploy this. I'm going to merge this into a `v1` branch in lsp2clj, to get it out of my personal fork, and so that the clojure-lsp branch can depend on a `:git/sha` until we have a release. I'll open another PR (essentially a duplicate of this one) asking to merge `v1` into `master`. We can do that merge after testing.

Cyrik:
Thanks for the great work @mainej. I've been on a business trip but I'll try to take a look at this next week.

PR #13, opened 2022-07-15
Remove lsp4j

### LINKED ISSUE #8: Replace lsp4j with internal communication logic

mainej:
Replaces lsp4j with a Clojure implementation. The full list of dependencies is now clojure, core.async, cheshire, and camel-snake-kebab.

Also merges the lsp4clj-server and lsp4clj-protocol projects into one top-level project, lsp4clj.

clojure-lsp has an [accompanying PR](https://github.com/clojure-lsp/clojure-lsp/pull/1117).

@Cyrik, as stated in the CHANGELOG, other users of lsp4clj (you, I think) are encouraged to upgrade. You can follow that clojure-lsp PR as a guide on the changes that'll need to be made. They're fairly extensive, but I believe will make lsp4clj a more stable platform for LSP development. When you're upgrading if you'd like a little direction, I'd be happy to chat. Perhaps we can turn our discussion into edits to the Usage section in the README.

Notable changes:

- As described in #8, servers can choose which requests and notifications they respond to by implementing the `lsp4clj.server/receive-request` and `lsp4clj.server/receive-notification` multimethods. (In contrast to the protocols based approach we had before.) This gives servers much more flexibility. They can implement new parts of the spec at their leisure, or according to the needs of their language. They can also choose when and how to implement asynchronicity. The tradeoff is that much of the `spec` checking and logging moves back to clojure-lsp and other servers. The servers also have to know and use the names of the LSP requests/notifications. Personally, I find `clojure-lsp.server` code to be more readable now, even though it has a few more responsibilities.
- Input maps are all coerced to kebab-case. The server receives whatever the client sent, which should follow the LSP spec. 
- Output specs still exist, but do only very lightly conformation. Of what conformation remains, most of it converts keywords to enum values. There are a few places where more complex conformation happens. Honestly, I'd like to remove almost all conformation. I think it will be easier if you can look up what to output based on the LSP spec, not the LSP spec, plus a bunch of reasoning about coercion.
- It is much less code to initialize a server. None of it is Java interop.
- We have a path to working on #1. We need code that converts a socket into a pair of input and output channels. Otherwise, much of the code can stay the same.

Fixes #8.


src/lsp4clj/json_rpc.clj
```diff
@@ -0,0 +1,115 @@
...
+
+(defn output-stream->output-chan
+  "Returns a channel which expects to have messages put on it. nil values are
+  not allowed. Serializes and writes the messages to the output. When the
+  channel is closed, closes the output.
+
+  Writes in a thread to avoid blocking a go block thread."
+  [^java.io.OutputStream output]
+  (let [messages (async/chan 1)]
+    (binding [*out* (io/writer output)]
```
snoe: Is this only bound so write-message can use `print`? Would it be better to pass writer to `write-message`?

snoe: `io/writer` should be managed with `with-open` - this might need to be done within the thread.

src/lsp4clj/json_rpc.clj
```diff
@@ -0,0 +1,115 @@
...
+
+(defn ^:private parse-header [line headers]
+  (let [[h v] (string/split line #":\s*" 2)]
+    (when-not (contains? #{"Content-Length" "Content-Type"} h)
+      (throw (ex-info "unexpected header" {:line line})))
+    (assoc headers h v)))
+
+(defn ^:private parse-charset [content-type]
+  (or (when content-type
+        (when-let [[_ charset] (re-find #"; charset=(.*)$" content-type)]
```
snoe: I think it might be a good idea (now or eventually) to use a content-type parser here. 
At least, I would check charset in a case-insensitive way.

src/lsp4clj/json_rpc.clj
```diff
@@ -0,0 +1,115 @@
...
+          (when (< new-read 0)
+            ;; TODO: return nil instead?
+            (throw (java.io.EOFException.)))
+          (recur (+ total-read new-read)))))
+    (String. ^bytes buffer ^String charset-s)))
+
+(defn ^:private parse-header [line headers]
+  (let [[h v] (string/split line #":\s*" 2)]
+    (when-not (contains? #{"Content-Length" "Content-Type"} h)
+      (throw (ex-info "unexpected header" {:line line})))
```
snoe: This seems unnecessarily constricting. Should unexpected headers simply be skipped in the spirit of the [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle)?

src/lsp4clj/json_rpc.clj
```diff
@@ -0,0 +1,115 @@
...
+   [camel-snake-kebab.extras :as cske]
+   [cheshire.core :as json]
+   [clojure.core.async :as async]
+   [clojure.java.io :as io]
+   [clojure.string :as string]))
+
+(set! *warn-on-reflection* true)
+
+(defn ^:private read-n-chars [^java.io.InputStream input content-length charset-s]
+  (let [buffer (byte-array content-length)]
```
snoe, line 15: I think BufferedReader might be able to replace this.

snoe (commented):
Up to you if you want to address anything. Really great job!

src/lsp4clj/json_rpc.clj
mainej: Good suggestions. I'll make those changes, and introduce some tests to make sure it's working.

src/lsp4clj/json_rpc.clj
mainej: Not a bad idea. I was going off lsp4j, which is [equally naive](https://github.com/eclipse/lsp4j/blob/5605ca82252833777dbef82ca13a403dbfb9d906/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/StreamMessageProducer.java#L161). I'm going to hold off on importing a real header parser, but I'll make the regex case-insensitive.

src/lsp4clj/json_rpc.clj
mainej: Yeah, I suppose we should skip unexpected headers. That's what lsp4j [does](https://github.com/eclipse/lsp4j/blob/5605ca82252833777dbef82ca13a403dbfb9d906/org.eclipse.lsp4j.jsonrpc/src/main/java/org/eclipse/lsp4j/jsonrpc/json/StreamMessageProducer.java#L148-L168). I wish the LSP spec had just said, "headers are like those for HTTP", but instead, they're [overly specific](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#headerPart).

src/lsp4clj/json_rpc.clj
mainej: Originally I tried to write all this code with a BufferedReader, but couldn't figure out how to do it right. I was having trouble both with the Reader part and the Buffered part.

Readers work at the character level, but we're down at the byte level. Content-Length is specified in bytes, and in order to correctly parse the message we must read that many bytes, exactly, and interpret them in the given charset. (I've renamed this function `read-n-bytes` to make that clearer.) So as we're reading the content, we can't use a Reader (chars), we have to use an InputStream (bytes).

If you want to work at a higher level (for example, to `(read-line reader)` instead of `(read-header-line input)`. You have to switch back and forth between a Reader wrapper and a lower-level InputStream. But that can cause its own problems. If you use Buffered wrappers you don't have control over how many characters/bytes are read off the underlying InputStream. If a buffered wrapper reads too many characters, they're lost. When you switch back to the InputStream, it won't see those lost characters, meaning the next message will be malformed.

I wouldn't be surprised if there's a Java Way to do this, perhaps with Pushback readers. But that's more advanced than I understand. I decided it'd be easier for me to reason about (though more verbose) if I stayed down at the level of individual bytes. 

Anyway, all that to say, I'd welcome advice about how to make this more readable.

mainej:
Thanks for the feedback @snoe! I've made or responded to your suggested changes. Let me know if there's anything else you'd like to see changed.

src/lsp4clj/json_rpc.clj
snoe: Ah, it looks like java 9+ has readNBytes https://docs.oracle.com/javase/9/docs/api/java/io/InputStream.html#readNBytes-byte:A-int-int- I can't remember what we're targeting.

Otherwise, this is fine.

snoe (approved):

src/lsp4clj/json_rpc.clj
mainej: Oh too bad... readNBytes would be handy. But, we're still [targeting](https://github.com/clojure-lsp/clojure-lsp/blob/aa46e433e9642c1368336130d1dd99cfc68402a8/.github/workflows/ci.yml#L56) Java 8.

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
 ## Unreleased
 
+## v1.0.0
+
+- Remove lsp4j, as per https://github.com/clojure-lsp/lsp4clj/issues/8
+This is essentially a rewrite of lsp4clj. Users of lsp4clj v0.4.3 and earlier
+are encouraged to upgrade. Bug fixes for these earlier versions will be
+considered, but the lsp4j-based version of this library will not receive
+long-term support. For an example of how to use the new version of the library,
+see https://github.com/clojure-lsp/clojure-lsp/pull/1117
```
ericdallo, line 12: Since the current version is alpha, I'm totally fine merging v1 into master soon after it's running fine on clojure-lsp

README.md
```diff
@@ -2,18 +2,126 @@
...
+    (recur)))
+```
+
+### Receive messages
+
+To receive messages from a client, lsp4clj defines a pair of multimethods, `lsp4clj.server/receive-notification` and `lsp4clj.server/receive-request` that dispatch on the method name (as defined by the LSP spec) of an incoming JSON-RPC message.
+
+Server implementors should create `defmethod`s for the messages they want to process. (Other methods will be logged and responded to with a generic "Method not found" response.)
+
+These `defmethod`s receive 3 arguments, the method name, a "context", and the `params` of the [JSON-RPC request or notification object](https://www.jsonrpc.org/specification#request_object). The keys of the params will have been converted (recursively) to kebab-case keywords. Read on for an explanation of what a "context" is and how to set it.
```
ericdallo, line 34: One thing I think we lost using multimethods is that we can't require that clients implement specific methods like we had on protocol, the LSP spec has some methods that all server should provide, they are required methods, like initialize, diagnostics, definition etc. I'm ok for now with this, we can think on how to make required some methods in the future

README.md
```diff
@@ -2,18 +2,126 @@
...
+
+### Send messages
+
+Servers also initiate their own requests and notifications to a client. To send a notification, call `lsp4clj.server/send-notification`.
+
+```clojure
+(->> {:message message
+      :type type
+      :extra extra}
+     (conform-or-log ::coercer/show-message)
```
ericdallo, line 60: Why require conform? can't lsp4clj do the conform on its side as it knows the method and to what conform?

README.md
```diff
@@ -2,18 +2,126 @@
...
+
+If the request is cancelled, future invocations of `deref` will return `:lsp4clj.server/cancelled`.
+
+Sends `$/cancelRequest` only once, though `lsp4clj.server/deref-or-cancel` or `future-cancel` can be called multiple times.
+
+### Start and stop a server
+
+The last step is to start the server you created earlier. Use `lsp4clj.server/start`. This method accepts two arguments, the server and a "context".
+
+The context should be `associative?`. Whatever you provide in the context will be passed as the second argument to the notification and request `defmethod`s you defined earlier. This is a convenient way to make components of your system available to those methods without definining global constants. Often the context will include the server itself so that you can initiate outbound requests and notifications in reaction to inbound messages. lsp4clj reserves the right to add its own data to the context, using keys namespaced with `:lsp4clj.server/...`.
```
ericdallo, line 86: 💯 👏🏻

README.md
```diff
@@ -2,18 +2,126 @@
...
+
+```clojure
+(lsp4clj.server/start server {:server server, :logger logger})
```
ericdallo: Maybe use another example here than passing the server itself to make it clear for people who don't know the components concept?
maybe pass logger and some other map like `:my-client-settings` or something like that

ericdallo: Just a nit though

src/lsp4clj/coercer.clj
```diff
@@ -0,0 +1,432 @@
...
+
+(s/def :show-message-request/action (s/keys :req-un [:show-message-request-action/title]))
+
+(s/def :show-message-request/actions (s/coll-of :show-message-request/action))
+
+(s/def ::show-message-request (s/keys :req-un [:show-message/type
+                                               :show-message/message]
+                                      :opt-un [:show-message-request/actions]))
+
+;; (def work-done-progress-kind-enum
```
ericdallo: why commented?

src/lsp4clj/json_rpc.clj
```diff
@@ -0,0 +1,127 @@
...
+  that are already strings."
+  [k]
+  (cond-> k (keyword? k) csk/->camelCaseString))
+
+(def ^:private write-lock (Object.))
+
+(defn ^:private write-message [^java.io.OutputStream output msg]
+  (let [content (json/generate-string (cske/transform-keys kw->camelCaseString msg))
+        content-bytes (.getBytes content "utf-8")]
+    (locking write-lock
```
ericdallo, line 56: There was some issues in the past on Clojure with the locking macro on Graalvm images, I think it was solved, but it's good to keep in mind that case during tests with graalvm, I do think integration tests cover that though

src/lsp4clj/json_rpc/messages.clj
```diff
@@ -0,0 +1,63 @@
...
+  (let [response (assoc base-message :id id)]
+    (if (and (map? result) (:error result))
+      (assoc response :error (:error result))
+      (assoc response :result result))))
+
+(def error-codes
+  {;; JSON-RPC errors
+   ;; Is it possible to respond if you have a parse error? How could you reply
+   ;; to a request if you weren't able to parse it?
+   :parse-error      {:code -32700, :message "Parse error"}
```
ericdallo, line 28: Good point, I always saw that on the spec as: I tried to process a method like `find-references`, but the clojure code had a parse-error, so I could return this, WDYT? 🤷🏻‍♂️

src/lsp4clj/json_rpc/messages.clj
```diff
@@ -0,0 +1,63 @@
...
+#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
+(defn work-done-progress [percentage message progress-token]
+  (let [percentage (int percentage)
+        progress {:kind (case percentage
+                          0 :begin
+                          100 :end
+                          :report)
+                  :title message
+                  :percentage percentage}]
+    {:token (or progress-token "clojure-lsp")
```
ericdallo:
```suggestion
    {:token (or progress-token "lsp4clj")
```
Or something more generic

ericdallo (approved):
Looks great code-wise, awesome work!

CHANGELOG.md
mainej: @ericdallo how do you want to test v1 with clojure-lsp? I think it's easiest to pin it to a `:git/sha` near the tip of v1, but we could release v1 to clojars since the project name is changing.

README.md
mainej: To be pedantic, you're not _required_ to implement every method on a protocol, the docs just say that you _should_. If you don't, you don't get any compile time errors, though you do get a runtime error if you try to call the method you didn't define.

But I understand what you're saying—a protocol is a stronger indication of which methods you're expected to implement.

This is just one of the drawbacks to the multimethods. They also don't give you anywhere to hang extra data about handlers. I can imagine wanting to flag certain messages as cancellable, give them validators to apply to their input or output, or more generally give them middleware style wrappers, etc.

I think the best way to think about this is that one of the typical responsibilities of a server is to route requests. The multimethods work for that, but reitit, Compojure, et al. address the fact that there's more you might want to do with routing.

On the other hand, I haven't run into anything on the clojure-lsp side yet where I really regret the simplicity of the multimethods.

So, I guess my feeling is that multimethods might not be the best design long-term, but I'm not sure I have enough information to commit to another design either.

I've vaguely considered expressing the "router" as a hashmap, keyed by method name, which you pass in to the server when you create it. The server could validate that certain keys were present. That would also give us lots of flexibility about what the values of the hashmap are. They could be functions to call as the handler (much like a multimethod, which maps dispatch values to functions). Or they could be other hashmaps, with a handler function in each. We could add other data over time. If you feel strongly that this is a better solution, I can work on it.

README.md
mainej: That's true, lsp4clj has enough information to do conformation automatically. I guess the main reason I didn't do it that way is that I don't like the conformation, and so I wanted to make it optional.

In the particular case of `::coercer/show-message`, the reason I don't like the conformation is that it is a very complicated way to convert `:type` from a keyword like `:warning` to an enum value like `2`. Otherwise, it serves no purpose, IMO. Beyond `::coercer/show-message`, I feel the same way about most of the conformation. To me, the conformation is a relic of lsp4j. It was a sane way to create Java objects, but it is too convoluted for doing simple translations like `:warning` -> `2`.

I started to abandon the conformation but then decided that it'd add too much noise to an already big change set. If you'd like, we can work on moving the conformation back to the server, but I'm inclined to do that in a separate PR.

This is another place where having a "route" object could be handy. You could add a `:route/conform-output?` flag which indicated to the server that it should do its default conformation. That would let clojure-lsp opt-out.

src/lsp4clj/coercer.clj
mainej: Progress notifications were one of the first things I implemented. Instead of doing conformation, I made `lsp4clj.json-rpc.messages/work-done-progress`, a function which does the same manipulation as this coercer code did—converting 0 to `:begin`, 100 to `:done` and other related transformations. I thought I would do that for all the output, replacing all the conformation. But that ended up being too much work.

So, this piece of the coercer got commented out, while other pieces got left in, although modified to remove the Java conversions. I'd be happy to uncomment it, if you'd like, but there won't be any users of it.

src/lsp4clj/json_rpc.clj
mainej: I read a little about that bug—yikes. Seems nasty. It sounds like it's fixed though. The GraalVM build works locally for me on a Mac, and as you said, if it were broken for Linux we'd see that in the integration tests.

src/lsp4clj/json_rpc/messages.clj
mainej: True, it's not super easy to tell. The JSON-RPC spec says that there are three levels of error, one for the JSON level, one for the server level (lsp4clj) and one for the application level (clojure-lsp). The spec says that these [specific error codes](https://www.jsonrpc.org/specification#error_object), -32700 and -32600 are for the lowest JSON layer. If the application layer has a parse error, I think it's supposed to return an error >= -32100. I don't really know though—it's not very carefully specified and I doubt anyone relies on the error codes ranges. In any case, I couldn't find a way to return them to the client, though we do return other predefined error codes at times.

CHANGELOG.md
ericdallo: I'm ok with using git/sha until we think it's stable enough to merge to master and release

README.md
ericdallo: > To be pedantic, you're not required to implement every method on a protocol, the docs just say that you should. If you don't, you don't get any compile time errors, though you do get a runtime error if you try to call the method you didn't define.

yeah, the issue is that most clients expect that some methods are always available like diagnostics, but I agree it's best to be a little bit more loose than strict for now.

> Or they could be other hashmaps, with a handler function in each. We could add other data over time. If you feel strongly that this is a better solution, I can work on it.

I think it would work well, but I'm ok trying the multimethods for now

README.md
ericdallo: I see, most of conformation is not needed anymore indeed, we can think in the future replacing with safe checks/adapters and only that probably, it'd make project easier to understand too I guess

src/lsp4clj/coercer.clj
ericdallo: I think we can remove it so?

src/lsp4clj/json_rpc/messages.clj
ericdallo: Yeah, makes sense

README.md
mainej: Yep, that's my hope. We'll see though... There's rarely any reason to change the existing specs, so they might stick around.

README.md
mainej: Changed

README.md
Cyrik: I'm also split on this. On the one hand, multimethods bring great simplicity, but our tooling for them is a lot worse. Previously I could find the implementation of the specific message handler with a "find reference", that's gone with multimethods. Additionally, there's no indication of all possible methods I could implement.
I still like the multimethod approach, but maybe this will drive us to add better multimethod stuff into clojure-lsp 😄

ericdallo: @Cyrik we improved a lot defmethods support on clojure-lsp, doesn't find implementations work for that case?

mainej: I'd say there are two parts to discovery in this case... What has been implemented? And what should be implemented?

For me, in Emacs, `find-implementation` works well for finding what has been implemented.

Screen Shot 2022-07-31 at 2 25 11 PM

In the last release we also improved the way defmethods appear in `document-symbols` (which editors use to make a sidebar with all the file's functions) so that's another way to figure out what's been implemented.

Screen Shot 2022-07-31 at 2 25 48 PM

But as @ericdallo mentioned originally, the multimethods leave something to be desired as far as answering what should be implemented. In my opinion, the protocols did a slightly better job of that, but only marginally. The protocols were essentially various lists of things that clojure-lsp had implemented, but required and optional features were mixed together. So, I'd argue that the true authority on what should be implemented was and still is the LSP spec.

But I think it'll be interesting to explore whether lsp4clj can improve that aspect of discoverability over time. Having the "routes" be a map of handlers would probably mean the editors would have even less tooling to help you find code. But lsp4clj could provide more feedback about whether your routes were valid.

Anyway, I'm planning to leave the multimethods as is, for now. IIRC pedestal went through 3 or 4 different router implementations before they finally found one that was a good mix of declarative, fast and flexible, and I think that's the one most projects use now. So, we're allowed to get it wrong several more times before we get it right!

Cyrik: The thing I'm missing from defmethods is a way of seeing references to a specific method. I know that this isn't possible in the general case, which is the whole point of defmethods, but often enough there are hardcoded calls to a specific method that could be shown as a direct reference. Something like `(receive-request "shutdown")`.
The other thing is completeness, at least in this case.

But in I also fully agree with @mainej that we should leave this as is for now since these are very minor gripes with the system.

mainej: @Cyrik let me see if I follow...

Suppose we have this multimethod:
```clojure
(defmulti caps identity)

(defmethod caps "a" [_] "A")
(defmethod caps "b" [_] "B")
```

At call sites like the following, you know for certain that `(defmethod caps "a" ...)` is being called, not `(defmethod caps "b" ...)`.
```clojure
(caps "a")
```
In theory `clojure-lsp` could provide navigation tools to link the `a` defmethod to its usage.

Of course this isn't possible when it's ambiguous which defmethod is being called:
```clojure
(let [lower (rand-nth ["a" "b"])]
  (caps lower))
```
In these situations `clojure-lsp` couldn't help.

Assuming I'm following, I agree, that makes it harder to understand how multimethods, or really any kind of polymorphism, is being used. Polymorphism limits tooling, and that's true of all editors and languages. Maybe the next generation of tooling will offer more insight into polymorphic calls!

For these particular `lsp4clj` multimethods—`receive-request` and `receive-notification`—I'd say the problem isn't very big. There's only one place that calls `receive-request` and one that calls `receive-notification` (both in `lsp4clj`), so there's not much need to find the call sites.

I think there's a stronger argument that the other main entrypoints into `lsp4clj`—`send-request` and `send-notification`—exhibit a similar problem, even though they aren't multimethods. They accept the JSON-RPC method as an argument, so it could be difficult to find all the places that, for example, call `(lsp.server/send-notification server "$/progress" message)`. In `clojure-lsp` we've worked around this. We have only one place that calls `$/progress`, a function called `publish-progress`. If we want to find all the call sites, we can use `find-references` on `publish-progress`. And in fact, all the calls to `send-request` and `send-notification` are in similar wrapper functions, so it's pretty easy to find references.

Cyrik: @mainej Yeah that's the exact case I'm talking about. In my experience, most multimethods are built in this way (using the first param for dispatch). But all of this is very much unrelated so we should probably stop the topic 😄 

To your second point: you're correct that this happens more often, I just didn't notice the change yet. Previously it was easy to find references on the `ILSPProducer` methods and you'll now have to do the work yourself. Maybe this could be pulled into lsp4clj as optional entry points. But I know that @ericdallo and I have differing opinions on how much lsp4clj should provide, so I'm also happy to just do that on our end. At some point, I want to publish a library with all the extra utility stuff around lsp4clj anyway (file-management, URL handling, and so on).

ericdallo: we could open a new discussion on lsp4clj to see if we could provide something that help you like a different project using lsp4clj under the hood, I think discussing the problem a little bit more would make us understand better the problem

mainej: https://github.com/clojure-lsp/lsp4clj/issues/14

PR #15, opened 2022-08-10
Fix coercion of resolved completion items

mainej:
Coerce an integer `CompletionItem.insertTextFormat`—which is provided as input to `completionItem/resolve`—into a keyword. Then the keyword can be coerced back to an integer later, when `completionItem/resolve` returns the resolved CompletionItem.


ericdallo (approved):

ericdallo:
Tagged as 1.0.1

mainej:
Thanks!

PR #16, opened 2022-08-13
Support communication through a socket

### LINKED ISSUE #1: Support for more communication channels

mainej:
Adds `lsp4clj.socket-server/socket-server`, as an alternative to `lsp4clj.server/stdio-server`.

```clojure
(lsp4clj.socket-server/socket-server {:port "61235"})
```

This starts a socket server listening on the provided port, blocking until a client makes a connection. When the connection is made it returns a lsp4clj server that has the same behavior as a `stdio-server`, except that messages are exchanged over the socket. When the server is shut down, the connection will be closed.

@SevereOverfl0w, I'm going to say that this fixes #1. However, it wasn't clear from #1 whether you were hoping to analyze Clojure code. If you were, then there's more to do. At the risk of explaining things you might already understand, this project, lsp4clj, is agnostic about languages. All of the tools for analyzing Clojure code are over in clojure-lsp. lsp4clj provides the low-level communication tools that clojure-lsp is built on, but doesn't have any smarts about Clojure itself. If you'd like clojure-lsp to be able to listen on a socket, then feel free to open an issue [over there](https://github.com/clojure-lsp/clojure-lsp/issues). This PR will be an important prerequisite to that work. If you just wanted tooling to implement an LSP server that listens on a socket, then hopefully this PR has what you need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,10 +97,10 @@ When a server shuts down it stops reading input, finishes processing the message
 
 ### Socket server
 
-The `stdio-server` is the most commonly used, but the library also provides a `lsp4clj.server/socket-server`.
+The `stdio-server` is the most commonly used, but the library also provides a `lsp4clj.socket-server/server`.
 
 ```clojure
-(lsp4clj.server/socket-server {:port 61235})
+(lsp4clj.socket-server/server {:port 61235})
 ```
 
 This will start listening on the provided port, blocking until a client makes a connection. When the connection is made it returns a lsp4clj server that has the same behavior as a `stdio-server`, except that messages are exchanged over the socket. When the server is shut down, the connection will be closed.
@@ -130,7 +130,7 @@ You may also find `lsp4clj.server/chan-server` a useful alternative to `stdio-se
 
 ## Caveats
 
-You must not print to stdout while a `stdio-server` is running. This will corrupt its output stream and clients will receive malformed messages. To protect a block of code from writing to stdout, wrap it with `lsp4clj.server/discarding-stdout`. The `receive-notification` and `receive-request` multimethods are already protected this way, but tasks started outside of these multimethods need this protection added. Consider using a `lsp4clj.server/socket-server` to avoid this problem.
+You must not print to stdout while a `stdio-server` is running. This will corrupt its output stream and clients will receive malformed messages. To protect a block of code from writing to stdout, wrap it with `lsp4clj.server/discarding-stdout`. The `receive-notification` and `receive-request` multimethods are already protected this way, but tasks started outside of these multimethods need this protection added. Consider using a `lsp4clj.socket-server/server` to avoid this problem.
 
 ## Known lsp4clj users
 
--- a/README.md
+++ b/README.md
@@ -73,11 +73,11 @@ Sending a request is similar, with `lsp4clj.server/send-request`. This method re
     response))
 ```
 
-Otherwise, the request object presents the same interface as `future`. Responds to `future-cancel` (which sends `$/cancelRequest`), `realized?`, `future?`, `future-done?` and `future-cancelled?`.
+The request object presents the same interface as `future`. It responds to `future-cancel` (which also sends `$/cancelRequest`), `realized?`, `future?`, `future-done?` and `future-cancelled?`.
 
-If the request is cancelled, future invocations of `deref` will return `:lsp4clj.server/cancelled`.
+If the request is cancelled, later invocations of `deref` will return `:lsp4clj.server/cancelled`.
 
-Sends `$/cancelRequest` only once, though `lsp4clj.server/deref-or-cancel` or `future-cancel` can be called multiple times.
+`$/cancelRequest` is sent only once, although `lsp4clj.server/deref-or-cancel` or `future-cancel` can be called multiple times.
 
 ### Start and stop a server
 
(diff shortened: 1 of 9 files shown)


SevereOverfl0w:
This sounds exactly like what I need, thanks!

Cyrik:
Great work @mainej!
Two thoughts on this though: 
The way this is implemented makes us more reliant on the JVM. I was hoping we could get to a CLJS version at some point, so it might make sense to pull in the JVM stuff optionally? Not sure if you and @ericdallo see this as a goal.

Secondly, I wanted the same thing for websockets, just to have a more portable means of communication. Would you want to do this here or should I start it as a separate PR at some point?

ericdallo:
Awesome work!
@Cyrik To be honest I didn't have the goal to offer support for cljs or websockets, since the LSP spec just mention stdio and sockets, so if we intend to support that, would be something custom, I don't see a problem with that, if we keep the code simple enough I'm ok with that.

README.md
```diff
@@ -114,13 +124,13 @@ As you are implementing, you may want to trace incoming and outgoing messages. I
...
 
 ## Caveats
 
-You must not print to stdout while a `stdio-server` is running. This will corrupt its output stream and clients will receive malformed messages. To protect a block of code from writing to stdout, wrap it with `lsp4clj.server/discarding-stdout`. The `receive-notification` and `receive-request` multimethods are already protected this way, but tasks started outside of these multimethods need this protection added. See https://github.com/clojure-lsp/lsp4clj/issues/1 for future work on avoiding this problem.
+You must not print to stdout while a `stdio-server` is running. This will corrupt its output stream and clients will receive malformed messages. To protect a block of code from writing to stdout, wrap it with `lsp4clj.server/discarding-stdout`. The `receive-notification` and `receive-request` multimethods are already protected this way, but tasks started outside of these multimethods need this protection added. Consider using a `lsp4clj.server/socket-server` to avoid this problem.
```
ericdallo: 👍🏻

src/lsp4clj/server.clj
```diff
@@ -245,3 +253,29 @@
...
+(defn init-socket-server [{:keys [address port]}]
+  (json-rpc/start-socket-server {:address address
+                                 :port (if (string? port) (parse-long port) port)}))
+
+#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
+(defn socket-server
+  "Starts listening on the address (loopback by default) and port. Blocks until
+  a client establishes a connection, then returns a chan-server which
+  communicates over the socket."
+  [{:keys [_address _port] :as opts}]
```
ericdallo: Any reason for this destructuring other than making explicit what opts receives?

test/lsp4clj/server_test.clj
```diff
@@ -440,3 +442,43 @@
...
+(deftest should-communicate-through-socket
+  (let [;; NOTE: In real-life, you'd replace the following two lines with
+        ;; (server/socket-server {:port 51234}). In tests, setup is split into
+        ;; two steps for two reasons.
+        ;; 1. We have to ensure the server is started before the client, to
+        ;;    avoid flakiness. If we had just `(future (server/socket-server
+        ;;    {:port 51234}))` occasionally the client would start before the
+        ;;    server, leading to a ConnectionRefused exception. The only other
+        ;;    way to avoid this flakiness is to sleep briefly, or to poll for
+        ;;    the socket to be isBound.
```
ericdallo: Good catch

ericdallo (approved):

src/lsp4clj/server.clj
mainej: No reason other than to be clear about the opts. The docstring mentions address and port, so I wanted to make the arg list match. I think the "correct" way to do this is to attach `:arglists` metadata, but I never do that because of a combination of laziness and fear that it'll get out of date. 😅

mainej:
> The way this is implemented makes us more reliant on the JVM. 

I know... I was reluctant to add a bunch of `:import` statements.

> I was hoping we could get to a CLJS version at some point, so it might make sense to pull in the JVM stuff optionally? Not sure if you and @ericdallo see this as a goal.

I can definitely imagine a CLJC version of lsp4clj. Fundamentally lsp4clj is all about reading and writing bytes in a specific format, so there's no reason it can't work in CLJS. But there's a lot to refactor to get to that point.

Even before this PR, `lsp4clj.json-rpc` was highly tied to the JVM, depending deeply on java.io.InputStream and java.io.OutputStream. And since the logic in that namespace is crucial to interpreting LSP's flavor of JSON-RPC, the whole project is still very tied to the JVM. So that whole namespace would have to be re-worked.

Anyway, I'd say 👍 on CLJS (it would be a nice story about the evolution of lsp4clj) but 👎 on making that part of this PR.

If we do that work eventually, I think it'd be nice to maintain the public interface of `lsp4clj.server`. I view the rest of the project as implementation details. We can refactor it however we like to make it amenable to being used from CLJS. And even if we have to change `lsp4clj.server`, that'd be OK. It's a small enough user base that I wouldn't feel terrible about changing the API.

> Secondly, I wanted the same thing for websockets, just to have a more portable means of communication. Would you want to do this here or should I start it as a separate PR at some point?

I can see some use cases for websockets. So yes, a separate PR would be great. Feel free to refactor the code in this PR however you like as you're doing that. I view this PR as a way to prove that _something_ will work, not as a final decision about how this code should look.

mainej:
Actually I'll make a few little changes to separate responsibilities a little. Might help support CLJS someday.

mainej:
@Cyrik I've moved things around so that the `java.net` dependencies are only pulled in if you require `lsp4clj.socket-server`. I think it's better this way—thanks for the suggestion.

In retrospect, I wish `lsp4clj.server/stdio-server` was in its own (very small) namespace too. Then the README could say that you pick an implementation namespace (`lsp4clj.stdio-server`, `lsp4clj.socket-server`, or someday `lsp4clj.websocket-server`), use that namespace to create a server, then interact with that server through `lsp4clj.server`.

src/lsp4clj/server.clj
mainej: The latest refactoring uses `:address` and `:port`, so the destructuring actually has a point now. :)

Cyrik:
nice refactor, thanks @mainej!

PR #17, opened 2022-08-15
Deprecate server/stdio-server

mainej:
This introduces a new namespace `lsp4clj.io-server`. It deprecates `lsp4clj.server/stdio-server` and suggests `lsp4clj.io-server/server`, which has identical behavior, as a replacement. It also introduces `lsp4clj.io-server/stdio-server`, which uses System/in and System/out by default, providing a smaller API for servers. Finally, it adds integration level tests for an io-server, something which we only had disconnected unit level test for before.

This is the first step in isolating the parts of the code base that use Java interop, a goal which was first observed in [discussions](https://github.com/clojure-lsp/lsp4clj/pull/16#issuecomment-1214206756) around #16. I have a follow-up commit which removes the deprecated function and merges `lsp4clj.json-rpc` (which has all the InputStream and OutputStream interop) into `io-server`. That will leave `lsp4clj.server` more (though not entirely) independent of Java. That should be part of a subsequent release so that we have at least one release with the deprecation warnings.

### DIFF
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -251,7 +251,7 @@
 (defn stdio-server
   "DEPRECATED: Will be removed in a future release. Prefer
   lsp4clj.io-server/server or lsp4clj.io-server/stdio-server."
-  {:deprecated true}
+  {:deprecated "Use lsp4clj.io-server/server instead"}
   [{:keys [in out] :as opts}]
   (chan-server (assoc opts
                       :input-ch (json-rpc/input-stream->input-chan in)
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -247,7 +247,11 @@
      :pending-requests* (atom {})
      :join (promise)}))
 
+#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (defn stdio-server
+  "DEPRECATED: Will be removed in a future release. Prefer
+  lsp4clj.io-server/server or lsp4clj.io-server/stdio-server."
+  {:deprecated true}
   [{:keys [in out] :as opts}]
   (chan-server (assoc opts
                       :input-ch (json-rpc/input-stream->input-chan in)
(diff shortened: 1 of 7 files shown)


src/lsp4clj/server.clj
```diff
@@ -247,7 +247,11 @@
      :pending-requests* (atom {})
      :join (promise)}))
 
+#_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (defn stdio-server
+  "DEPRECATED: Will be removed in a future release. Prefer
+  lsp4clj.io-server/server or lsp4clj.io-server/stdio-server."
+  {:deprecated true}
```
ericdallo: I'm not sure what is the standard here, but usually I add a string to the value fo the deprecated keyword like:
```suggestion
  {:deprecated "Use lsp4clj.io-server/server instead"}
```

test/lsp4clj/io_server_test.clj
```diff
@@ -0,0 +1,39 @@
...
+   [clojure.test :refer [deftest is]]
+   [lsp4clj.io-server :as io-server]
+   [lsp4clj.json-rpc :as json-rpc]
+   [lsp4clj.json-rpc.messages :as messages]
+   [lsp4clj.server :as server]
+   [lsp4clj.test-helper :as h]))
+
+(set! *warn-on-reflection* true)
+
+(deftest server-test
```
ericdallo, line 13: Nice!

ericdallo (approved):
Looks good!
Not related to your change, but I think would be nice to have those API ns documented on cljdocs similar to how we have on clojure-lsp's readme pointing to https://cljdoc.org/d/com.github.clojure-lsp/clojure-lsp/2022.07.24-18.25.43/doc/readme

mainej:
@ericdallo agreed, docs could be better. Do you mean we should add a link to the [lsp4clj cljdocs](https://cljdoc.org/d/com.github.clojure-lsp/lsp4clj/1.0.1/doc/readme) somewhere? Or add docstrings to the ns forms? Or docstrings to the main functions? Probably all 3 would be good.

ericdallo:
All 3 would be good, but having lsp4clj main ns on cljdocs would be the best IMo

PR #18, opened 2022-08-16
Co-locate IO dependencies

mainej:
This deletes the deprecated function `lsp4clj.server/stdio-server`, and merges `lsp4clj.json-rpc` into `lsp4clj.io-server`, completing the refactoring described in #17. This reduces (though does not remove) `lsp4clj.server`'s dependencies on Java classes.


mainej:
I'm taking this out of draft. The deprecation in #17 was released as part of 1.1.0, so there's one version that includes the warning. clojure-lsp was updated in https://github.com/clojure-lsp/clojure-lsp/pull/1200.

~Note that this PR deletes the lsp4clj.json-rpc.messages namespace, without a deprecation warning. I could revert that deletion and add a deprecation warning, although I'm inclined to just fix clojure-lsp along with the next upgrade.~ Edit: the removal of lsp4clj.json-rpc.messages is actually in another PR I haven't submitted yet.

ericdallo (approved):

ericdallo:
Thanks!

PR #19, opened 2022-08-17
Introduce a helper to preserve kebab-case

mainej:
This mirrors the helper introduced in https://github.com/clojure-lsp/clojure-lsp/pull/1198. Eventually the helper in clojure-lsp should be removed and replaced with this one, so that clojure-lsp doesn't need a direct dependency on camel-snake-kebab.


ericdallo (approved):

ericdallo:
Tagged as 1.1.0

PR #28, opened 2022-10-05
Improve tracing

### LINKED ISSUE #27: Trace level should be configurable
This is two related feature requests.

First, it would be nice if it were possible to control the verbosity of the server traces. Currently, they are either on or off. But it would be nice to have an intermediate level, where the time/direction/method/id were logged, but the body/params/etc. were NOT.

Second, it would be nice to be able to control the trace level at runtime, so that traces could be turned on momentarily.

The LSP spec has outlined how to do both of these things.

It defines [TraceLevel](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#traceValue), an enum of 'off', 'messages' and 'verbose'. (I think our current traces map to 'off' and 'verbose', so I would like to add 'messages'.)

The initial level can be set in the [InitializationParams](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams) as `trace`. During the lifetime of a server the TraceLevel can be changed with a [$/setTrace](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#setTrace) notification. (Calva sends this notification when you change the trace level in the Calva settings window, though lsp4clj and clojure-lsp ignore it. I think the intention is to have a single setting that controls both the client and server trace level. I haven't noticed the notification from other editors, though they may send it too.)

I'm starting this Issue here in lsp4cl. It will need to generate traces that respect the trace level. And it will need some way of being told what the trace level should be. But clojure-lsp will have some responsibility too... I think it should handle the `initialize` and `$/setTrace` messages, and instruct lsp4clj what to do. I think it's better that the language servers handle these messages, because they may want to make their own decisions both about lsp4clj's trace level, but also about their own logging.

1. Ensure server is initialized with a trace-ch, to be prepared for level changes
2. Initialize server with a (stateful) trace level, defaulting to 'off'
3. Skip tracing if trace level is 'off`. Modify trace contents based on whether level is 'messages' or 'verbose'
4. Add `lsp4clj.protocols.endpoint.IEndpoint/set-trace-level`, and use it to modify trace level
5. (clojure-lsp) Call `set-trace-level` during `initiailize`
6. (clojure-lsp) Call `set-trace-level` upon `$/setTrace`

The new `--trace` option in clojure-lsp may still be useful, to control tracing _before_ the `initialize` message.

mainej:
Support tracing at an intermediate level, without message body details. And let language servers change trace level on running servers.

Fixes #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,16 +111,20 @@ This will start listening on the provided port, blocking until a client makes a 
 
 ### Tracing
 
-As you are implementing, you may want to trace incoming and outgoing messages. Initialize the server with `:trace? true` and then read traces (two element vectors, beginning with the log level `:debug` and ending with a string, the trace itself) off its `:trace-ch`.
+As you are implementing, you may want to trace incoming and outgoing messages. Initialize the server with `:trace-level "verbose"` and then read traces (two element vectors, beginning with the log level `:debug` and ending with a string, the trace itself) off its `:trace-ch`.
 
 ```clojure
-(let [server (lsp4clj.io-server/stdio-server {:trace? true})]
+(let [server (lsp4clj.io-server/stdio-server {:trace-level "verbose"})]
   (async/go-loop []
     (when-let [[level trace] (async/<! (:trace-ch server))]
       (logger/log level trace)
       (recur)))
   (lsp4clj.server/start server context))
 ```
+
+`:trace-level` can be set to `"off"` (no tracing), `"messages"` (to show just the message time, method, id and direction), or `"verbose"` (to also show details of the message body).
+
+The trace level can be changed during the life of a server by calling, for example, `(ls4clj.server/set-trace-level server "messages")`. This can be used to respect a trace level received at runtime, either in an [initialize](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams) request or a [$/setTrace](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#setTrace) notification.
 
 ### Testing
 
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -157,7 +157,7 @@
 
 (defn trace [{:keys [tracer* trace-ch]} trace-f & params]
   (when-let [trace-body (apply trace-f @tracer* params)]
-    (async/put! trace-ch trace-body)))
+    (async/put! trace-ch [:debug trace-body])))
 
 ;; TODO: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize
 ;; * receive-request should return error until initialize request is received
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -111,17 +111,21 @@ This will start listening on the provided port, blocking until a client makes a
...
-(let [server (lsp4clj.io-server/stdio-server {:trace? true})]
+(let [server (lsp4clj.io-server/stdio-server {:trace-level "verbose"})]
   (async/go-loop []
     (when-let [[level trace] (async/<! (:trace-ch server))]
       (logger/log level trace)
       (recur)))
   (lsp4clj.server/start server context))
 ```
 
+`:trace-level` can be set to `"off"` (no tracing), `"messages"` (to show just the message time, method, id and direction), or `"verbose"` (to also show details of the message body).
```
ericdallo, line 125: nit: maybe the value could be a keyword? `:messages`, `:off` etc

README.md
```diff
@@ -111,17 +111,21 @@ This will start listening on the provided port, blocking until a client makes a
...
+
+The trace level can be changed during the life of a server by calling, for example, `(ls4clj.server/set-trace-level server "messages")`. This can be used to respect a trace level received at runtime, either in an [initialize](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeParams) request or a [$/setTrace](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#setTrace) notification.
```
ericdallo, line 127: awesome!

README.md
mainej: My thinking was that keywords would be more Clojure-y, but since the spec defines these 3 values as strings, that’s what language servers will get from clients. It seems like an extra step to make them `(keyword trace)`. I’d be ok with either—what would you prefer?

ericdallo: Makes sense, I think string would be better indeed

mainej:
@ericdallo I'm going to refactor the `trace` ns a little. Please hold off on merging. Or would you prefer if I merge when I'm ready?

src/lsp4clj/server.clj
```diff
@@ -207,32 +213,32 @@
           now (.instant clock)
           req (lsp.requests/request id method body)
           pending-request (pending-request id method now this)]
-      (some-> trace-ch (async/put! (trace/sending-request req now)))
+      (trace this trace/sending-request req now)
```
ericdallo, line 216: 👍🏻

ericdallo (approved):
Looks great!

ericdallo:
Feel free to merge or do a new commit (Avoid force push if possible so I can review only the new commits)

mainej:
I'm going to go ahead and merge, but the commit is 52df404ca2f6e18ccf9d9aa55b0ba959db05b60b if you're interested. After you've done a release (doesn't have to be soon) will you ping me so I can start working on the clojure-lsp side?

ericdallo:
Tagged as 1.4.0 @mainej

PR #30, opened 2022-10-07
Let language servers abort cancelled running requests

mainej:
See the README for an explanation for why this is necessary. clojure-lsp needs it so that it can abort process-after-changes requests, which otherwise will run even if they are cancelled.


ericdallo (approved):
Nice docs!

PR #34, opened 2023-01-07
Buffer I/O

mainej:
Reduce latency introduced by ***lsp4clj*** by using buffered input and output streams.

In theory, it's faster to read data off the underlying stream in chunks, even though sometimes we need to process them one byte at a time. In practice the difference is probably minimal, though it doesn't hurt.

This uses `clojure.java.io/input-stream` and `output-stream` to create the buffered streams, which has the side-benefit of cleaning up the `socket-server` code slightly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Buffer I/O, slightly reducing latency.
+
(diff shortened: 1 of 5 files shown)


ericdallo (approved):
Looks great!

PR #35, opened 2023-01-28
Fix progress messages in Calva

mainej:
The helper to generate progress messages was incorrectly using the :title key in :report messages. That key should only be used with :begin messages. This changes the helper to generate valid progress messages.

After lsp4clj is bumped in clojure-lsp, Calva will show progress messages during startup. Emacs was showing progress messages before this fix, so it must observe the non-standard :title key.

I've also noticed that Calva doesn't show percentages in its UI. So, as a future enhancement to clojure-lsp, we may want to embed percentages in the messages themselves "36%", or perhaps add something like "file 26/235".

### DIFF
--- a/src/lsp4clj/lsp/requests.clj
+++ b/src/lsp4clj/lsp/requests.clj
@@ -16,6 +16,9 @@
 ;; TODO: The following are helpers used by servers, but not by lsp4clj itself.
 ;; Perhaps they belong in a utils namespace.
 
+(defn clamp [n n-min n-max]
+  (-> n (max n-min) (min n-max)))
+
 #_{:clj-kondo/ignore [:clojure-lsp/unused-public-var]}
 (defn work-done-progress
   "Returns the params for a WorkDone $/progress notification.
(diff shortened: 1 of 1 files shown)


ericdallo:
Oh, that's awesome, I will test that.
I thought that calva issue was because of [this](https://github.com/microsoft/vscode-languageserver-node/issues/828) and [this](https://github.com/clojure-lsp/clojure-lsp/issues/1363), didn't know we could fix only via that way

ericdallo:
Works great! thanks

ericdallo:
About the percentage, I'd like to avoid changing the message since will affect other clients (unless we add a flag just for calva wichi is kind of meh), maybe we could debug calva vscode-language-server-node to understand why the percentage is not showing

mainej:
> maybe we could debug calva vscode-language-server-node to understand why the percentage is not showing

Did some research... users have to choose to see the percentage. VS Code omits the percentage from the "window" progress area (the lower left), as part of a feature it calls "silent" notifications. It will show the progress message with the percentage in the "notification center", if it's open. There are several ways to open the notification center: by clicking on the progress messages, by clicking on the bell icon in the lower right, or by executing _Notifications: Show Notifications_.

ericdallo:
Got it, thanks @mainej maybe worth mentioning that in calva docs or tell to calva maintainers? It seems something that some users may like

PR #36, opened 2023-02-22
Let requests return errors

mainej:
According to the spec, all requests are allowed to return errors. This adds coercers to permit that.


ericdallo (approved):
I missed that recently, thank you!

PR #37, opened 2023-03-11
Continue receiving responses while blocking requests or notifications

### LINKED ISSUE #1500: Restart action on classpath fail message doesn't work
It seems that this [window/ShowMessageRequest](https://github.com/clojure-lsp/clojure-lsp/blob/c04df1c50afa5738e3e84f6367b1875ad84c3c8b/lib/src/clojure_lsp/classpath.clj#L131) is not working properly, the value is never returned until the `::timeout`, I did some debugging and testing with the reply the code works perfectly, so maybe it's related with that not working only during initialize?
Happens with Emacs/lsp-mode and VScode/Calva, so seems a server bug indeed.

mainej:
Fixes https://github.com/clojure-lsp/clojure-lsp/issues/1500.

I tested with the scenario described in https://github.com/BetterThanTomorrow/calva/issues/2067 and retry is now working. I'm not totally confident this is the most idiomatic way to solve this problem, but it's the best I've been able to come up with.

### DIFF
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -112,22 +112,79 @@
         log-title (format-error-code "Error receiving message" :internal-error)]
     (protocols.endpoint/log server :error e (str log-title "\n" message-details))))
 
-(defn ^:private receive-message
-  [server context message]
-  (let [message-type (coercer/input-message-type message)]
-    (try
+(defn ^:private spawn-receipt-thread [buf-or-n f]
+  (let [receipt-ch (async/chan buf-or-n)]
+    (async/thread
       (discarding-stdout
-        (case message-type
-          (:parse-error :invalid-request)
-          (protocols.endpoint/log server :error (format-error-code "Error reading message" message-type))
-          :request
-          (protocols.endpoint/receive-request server context message)
-          (:response.result :response.error)
-          (protocols.endpoint/receive-response server message)
-          :notification
-          (protocols.endpoint/receive-notification server context message)))
-      (catch Throwable e ;; exceptions thrown by receive-response or receive-notification (receive-request catches its own exceptions)
-        (log-error-receiving server e message)))))
+        (loop []
+          (when-let [[message-type message] (async/<!! receipt-ch)]
+            (f message-type message)
+            (recur)))))
+    receipt-ch))
+
+(defn ^:private run-pipeline
+  "Forwards messages received on the input-ch to the language server, for
+  further processing."
+  [server context input-ch]
+  (let [;; In order to process some requests and (all) notifications in series,
+        ;; the language server sometimes needs to block client-initiated input.
+        ;; If the language server sends requests during that time, it needs to
+        ;; receive responses, even though it's blocking other input. Otherwise,
+        ;; it will end up in a deadlock, where it's waiting to receive a
+        ;; response off the input-ch and the input-ch isn't being read from
+        ;; because the server is blocking input. See
+        ;; https://github.com/clojure-lsp/clojure-lsp/issues/1500.
+
+        ;; The messages all arrive in order on the input-ch so to get to the
+        ;; client's response, we have to queue whatever other messages it's
+        ;; sent. We do that by storing them in a sliding buffer. Because of the
+        ;; sliding buffer:
+        ;; * if the client sends a message which causes the language server to
+        ;;   block, and
+        ;; * if the language server sends a request during that time, and
+        ;; * if the client sends more than 100 other messages between when the
+        ;;   language server started blocking and when the client responds to
+        ;;   the language server's request,
+        ;; * then the client's earliest messages will be dropped.
+        ;; The same is true in reverse.
+
+        ;; We process the client- and language-server-initiated messages in
+        ;; separate threads.
+        ;; * Threads, so the language server can use >!! and so that we can use
+        ;;   (>!! output-ch) to respect back pressure from clients that are slow
+        ;;   to read.
+        ;; * Separate, so one can continue while the other is blocked.
+        server-initiated-in-ch (spawn-receipt-thread
+                                 (async/sliding-buffer 100)
+                                 (fn [_ message]
+                                   (try
+                                     (protocols.endpoint/receive-response server message)
+                                     (catch Throwable e
+                                       (log-error-receiving server e message)))))
+        client-initiated-in-ch (spawn-receipt-thread
+                                 (async/sliding-buffer 100)
+                                 (fn [message-type message]
+                                   (if (identical? :request message-type)
+                                       ;; receive-request catches its own exceptions
+                                     (protocols.endpoint/receive-request server context message)
+                                     (try
+                                       (protocols.endpoint/receive-notification server context message)
+                                       (catch Throwable e
+                                         (log-error-receiving server e message))))))]
+    (async/go-loop []
+      (if-let [message (async/<! input-ch)]
+        (let [message-type (coercer/input-message-type message)]
+          (case message-type
+            (:parse-error :invalid-request)
+            (protocols.endpoint/log server :error (format-error-code "Error reading message" message-type))
+            (:response.result :response.error)
+            (async/>! server-initiated-in-ch [:response message])
+            (:request :notification)
+            (async/>! client-initiated-in-ch [message-type message]))
+          (recur))
+        (do
+          (async/close! server-initiated-in-ch)
+          (async/close! client-initiated-in-ch))))))
 
 ;; Expose endpoint methods to language servers
 
@@ -199,21 +256,16 @@
                        join]
   protocols.endpoint/IEndpoint
   (start [this context]
-    (let [;; a thread so language server can use >!! and so that receive-message
-          ;; can use (>!! output-ch) to respect back pressure from clients that
-          ;; are slow to read.
-          pipeline (async/thread
-                     (loop []
-                       (if-let [message (async/<!! input-ch)]
-                         (do
-                           (receive-message this context message)
-                           (recur))
-                         (async/close! output-ch))))]
+    ;; Start receiving messages.
+    (let [pipeline (run-pipeline this context input-ch)]
+      ;; Wait to stop receiving messages.
       (async/go
-        ;; Wait for pipeline to close. This indicates input-ch was closed and
-        ;; that now output-ch is closed.
+        ;; When pipeline closes, it indicates input-ch has closed. We're done
+        ;; receiving.
         (async/<! pipeline)
-        ;; Do additional cleanup.
+        ;; Do cleanup.
+        ;; TODO: do we really know that we've finished putting to output-ch?
+        (async/close! output-ch)
         (async/close! log-ch)
         (some-> trace-ch async/close!)
         (on-close)
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -112,19 +112,19 @@
         log-title (format-error-code "Error receiving message" :internal-error)]
     (protocols.endpoint/log server :error e (str log-title "\n" message-details))))
 
-(defn ^:private spawn-receipt-thread [buf-or-n f]
-  (let [receipt-ch (async/chan buf-or-n)]
+(defn thread-loop [buf-or-n f]
+  (let [ch (async/chan buf-or-n)]
     (async/thread
       (discarding-stdout
         (loop []
-          (when-let [[message-type message] (async/<!! receipt-ch)]
-            (f message-type message)
+          (when-let [arg (async/<!! ch)]
+            (f arg)
             (recur)))))
-    receipt-ch))
+    ch))
 
-(defn ^:private run-pipeline
-  "Forwards messages received on the input-ch to the language server, for
-  further processing."
+(defn ^:private dispatch-input
+  "Dispatches messages received on the input-ch based on message type. Returns a
+  channel which will close after the input-ch is closed."
   [server context input-ch]
   (let [;; In order to process some requests and (all) notifications in series,
         ;; the language server sometimes needs to block client-initiated input.
@@ -154,23 +154,25 @@
         ;;   (>!! output-ch) to respect back pressure from clients that are slow
         ;;   to read.
         ;; * Separate, so one can continue while the other is blocked.
-        server-initiated-in-ch (spawn-receipt-thread
+
+        ;; (Jacob Maine): 100 is picked out of thin air. I have no idea how to
+        ;; estimate how big the buffer should be to avoid dropping messages. LSP
+        ;; communication tends to be very quiet, then very chatty, so it depends
+        ;; a lot on what the client and server are doing. I also don't know how
+        ;; many messages we could store without running into memory problems,
+        ;; since this is dependent on so many variables, not just the size of
+        ;; the JVM's memory, but also the size of the messages, which can be
+        ;; anywhere from a few bytes to megabytes.
+        server-initiated-in-ch (thread-loop
                                  (async/sliding-buffer 100)
-                                 (fn [_ message]
-                                   (try
-                                     (protocols.endpoint/receive-response server message)
-                                     (catch Throwable e
-                                       (log-error-receiving server e message)))))
-        client-initiated-in-ch (spawn-receipt-thread
+                                 (fn [response]
+                                   (protocols.endpoint/receive-response server response)))
+        client-initiated-in-ch (thread-loop
                                  (async/sliding-buffer 100)
-                                 (fn [message-type message]
+                                 (fn [[message-type message]]
                                    (if (identical? :request message-type)
-                                       ;; receive-request catches its own exceptions
                                      (protocols.endpoint/receive-request server context message)
-                                     (try
-                                       (protocols.endpoint/receive-notification server context message)
-                                       (catch Throwable e
-                                         (log-error-receiving server e message))))))]
+                                     (protocols.endpoint/receive-notification server context message))))]
     (async/go-loop []
       (if-let [message (async/<! input-ch)]
         (let [message-type (coercer/input-message-type message)]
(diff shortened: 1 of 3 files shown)


ericdallo:
Awesome! I'll take a closer look today and manually test, thanks in advance!

src/lsp4clj/server.clj
```diff
@@ -112,22 +112,79 @@
...
+        ;; * Separate, so one can continue while the other is blocked.
+        server-initiated-in-ch (spawn-receipt-thread
+                                 (async/sliding-buffer 100)
+                                 (fn [_ message]
+                                   (try
+                                     (protocols.endpoint/receive-response server message)
+                                     (catch Throwable e
+                                       (log-error-receiving server e message)))))
+        client-initiated-in-ch (spawn-receipt-thread
+                                 (async/sliding-buffer 100)
```
mainej, line 171: 100 is picked out of thin air. I have no idea how to estimate how big the buffer should be to avoid dropping messages. LSP communication tends to be very quiet, then very chatty, so it depends a lot on what the client and server are doing. I also don't know how many messages we could store without running into memory problems, since this is dependent on so many variables, not just the size of the JVM's memory, but also the size of the messages, which can be anywhere from a few bytes to megabytes.

src/lsp4clj/server.clj
```diff
@@ -112,22 +112,79 @@
...
+                                 (fn [message-type message]
+                                   (if (identical? :request message-type)
+                                       ;; receive-request catches its own exceptions
+                                     (protocols.endpoint/receive-request server context message)
+                                     (try
+                                       (protocols.endpoint/receive-notification server context message)
+                                       (catch Throwable e
+                                         (log-error-receiving server e message))))))]
```
mainej: I'm tempted to move the error logging here and around `receive-response` into the ChanServer. There was a point where it was cleaner to have the error handling "outside", but the way the code has evolved, that's no longer so.

src/lsp4clj/server.clj
```diff
@@ -112,22 +112,79 @@
...
+    (async/go-loop []
+      (if-let [message (async/<! input-ch)]
+        (let [message-type (coercer/input-message-type message)]
+          (case message-type
+            (:parse-error :invalid-request)
+            (protocols.endpoint/log server :error (format-error-code "Error reading message" message-type))
```
mainej, line 181: This is a separate issue, but I'd like to deprecate the `log` method on the `IEndpoint` protocol, replacing it with a private function in lsp4clj.server, much like `lsp4clj.server/trace`. It makes sense as a protocol method only if others need to _tell_ the server to log something. But neither clojure-lsp nor Scarlet have any need for that.

src/lsp4clj/server.clj
```diff
@@ -199,21 +256,16 @@
...
       (async/go
-        ;; Wait for pipeline to close. This indicates input-ch was closed and
-        ;; that now output-ch is closed.
+        ;; When pipeline closes, it indicates input-ch has closed. We're done
+        ;; receiving.
         (async/<! pipeline)
-        ;; Do additional cleanup.
+        ;; Do cleanup.
+        ;; TODO: do we really know that we've finished putting to output-ch?
+        (async/close! output-ch)
```
mainej, line 290: The [docs](https://clojuredocs.org/clojure.core.async/close!) for `close!` say A) "The channel will no longer accept any puts", B) "Data in the channel remains available for taking", and C) "Logically closing happens after all puts have been delivered."

At this point the input-ch has been closed, which means any messages that were read before the channel was closed have been put on the channel (C). However, the takes off of it, the takes which then forward the messages to the language server, may or may not have happened (B). And even if the language server has received some messages, if it responds after this line closes the output-ch, the responses will be dropped (A).

All that to say, I think it's possible for the lsp4clj server to drop the language server's final few responses.

It doesn't really matter though, because neither clojure-lsp nor Scarlet call `shutdown` on the lsp4clj server until they've received the `exit` notification, which is the client indicating it no longer expects any responses anyway.

test/lsp4clj/server_test.clj
```diff
@@ -133,6 +133,28 @@
...
+        client-resp (promise)]
+    (server/start server nil)
+    (with-redefs [server/receive-notification (fn [& _args]
+                                                (let [req (server/send-request server "server-sent-request" {:body "foo"})
+                                                      _ (Thread/sleep 100)
+                                                      resp (server/deref-or-cancel req 1000 :broke-deadlock)]
+                                                  (deliver client-resp resp)))]
+      (async/put! input-ch (lsp.requests/notification "client-sent-notif" {:input 1}))
+      (async/put! input-ch (lsp.requests/notification "client-sent-notif" {:input 2}))
+      (async/put! input-ch (lsp.requests/notification "client-sent-notif" {:input 3}))
```
mainej, line 160: My first pass at this, which used a `(chan 1)` instead of `(sliding-buffer 100)` worked when the client sent only two messages before the server could finish the first, but not when it sent 3 or more. The first notif made the server start sleeping, the second filled up the channel's buffer, and the third blocked on putting onto the channel, meaning the client response on the next few lines never got through. So these lines check that "several" client messages can be queued.

mainej (commented):
Here's some additional context which might be useful for the review. ~I'll move~ I've moved some of these into code comments.

src/lsp4clj/server.clj
```diff
@@ -199,21 +258,36 @@
...
+
+        ;; At this point the input-ch has been closed, which means any messages
+        ;; that were read before the channel was closed have been put on the
+        ;; channel (C). However, the takes off of it, the takes which then
+        ;; forward the messages to the language server, may or may not have
+        ;; happened (B). And even if the language server has received some
+        ;; messages, if it responds after this line closes the output-ch, the
+        ;; responses will be dropped (A).
+
+        ;; All that to say, it's possible for the lsp4clj server to drop the
```
ericdallo, line 283: Totally ok from my POV, the spec is [pretty clear](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#shutdown) that client should not send any other request than `exit`, so we could summarize as well as "we ignore any other client request other than exit as mentioned in the spec"

src/lsp4clj/server.clj
```diff
@@ -112,22 +112,81 @@
...
+        ;; If the language server sends requests during that time, it needs to
+        ;; receive responses, even though it's blocking other input. Otherwise,
+        ;; it will end up in a deadlock, where it's waiting to receive a
+        ;; response off the input-ch and the input-ch isn't being read from
+        ;; because the server is blocking input. See
+        ;; https://github.com/clojure-lsp/clojure-lsp/issues/1500.
+
+        ;; The messages all arrive in order on the input-ch so to get to the
+        ;; client's response, we have to queue whatever other messages it's
+        ;; sent. We do that by storing them in a sliding buffer. Because of the
```
ericdallo, line 140: Nice usage of sliding-buffer

src/lsp4clj/server.clj
```diff
@@ -112,22 +112,81 @@
...
+        ;; The same is true in reverse.
+
+        ;; We process the client- and language-server-initiated messages in
+        ;; separate threads.
+        ;; * Threads, so the language server can use >!! and so that we can use
+        ;;   (>!! output-ch) to respect back pressure from clients that are slow
+        ;;   to read.
+        ;; * Separate, so one can continue while the other is blocked.
+
+        ;; (Jacob Maine): 100 is picked out of thin air. I have no idea how to
```
ericdallo, line 158: I think we are safe with 100 messages, I can't see a case where makes sense a buffer bigger than that

ericdallo (approved):
Tested manually and it works! I found a minor problem with clojure-lsp settings hot-reload during this test, but not related and I have a fix coming to master.
Thank you for the help!

src/lsp4clj/server.clj
```diff
@@ -112,22 +112,81 @@
...
+        ;; client's response, we have to queue whatever other messages it's
+        ;; sent. We do that by storing them in a sliding buffer. Because of the
+        ;; sliding buffer:
+        ;; * if the client sends a message which causes the language server to
+        ;;   block, and
+        ;; * if the language server sends a request during that time, and
+        ;; * if the client sends more than 100 other messages between when the
+        ;;   language server started blocking and when the client responds to
+        ;;   the language server's request,
+        ;; * then the client's earliest messages will be dropped.
```
ferdinand-beyer, line 148: I think silently dropping messages could lead to confusing and hard to debug behaviour. Since both internal input channels use sliding buffers, we also effectively stop asserting back pressure on the client.

Instead of dropping messages, we could:

- Have multiple readers for `client-initiated-ch`, so that threads waiting for responses do not block the entire pipeline. This can currently be done in user-land by returning a promise from request handlers, maybe it can be generalised? Then again, this seems to just "move" the problem, as we could have all threads blocked eventually.

- Detect that buffers are full and raise an error, so that any server threads waiting for responses get unblocked. This is probably a bit tricky to implement, potentially using `offer!` (or `alts!`/`alt!`) and replacing the promise in `PendingRequest` with a promesa `deferred` that can be rejected for this purpose. This situation is similar to a timeout: The client did not produce a response "soon" enough.

I think I prefer the second approach, as there are no surprises (no deadlock, no silent message drops), and it seems to me that this kind of deadlock can only be prevented in handler code anyway. When a handler waits for a response, it needs to consider the situation that the response will never come.

By the way, we should also clear any `pending-sent-requests*` on shutdown anyway, see my other comment.

src/lsp4clj/server.clj
```diff
@@ -199,21 +258,36 @@
...
+    (let [pipeline (dispatch-input this context input-ch)]
+      ;; Wait to stop receiving messages.
       (async/go
-        ;; Wait for pipeline to close. This indicates input-ch was closed and
-        ;; that now output-ch is closed.
+        ;; When pipeline closes, it indicates input-ch has closed. We're done
+        ;; receiving.
         (async/<! pipeline)
-        ;; Do additional cleanup.
+        ;; Do cleanup.
```
ferdinand-beyer, line 268: Any threads blocked indefinitely waiting for a response will never get unblocked again.

I think we should handle any `pending-sent-requests*` in this case, unblocking waiters with an error. We could implement that by putting a "poison pill` on the internal input channels, as right here and now we do not know if the response is still buffered.

test/lsp4clj/server_test.clj
```diff
@@ -133,6 +133,37 @@
     (is (nil? (server/send-notification server "req" {:body "foo"})))
     (server/shutdown server)))
 
+(deftest should-receive-receive-response-to-request-sent-while-processing-notification
```
ferdinand-beyer, line 136: The double `receive` is probably a typo?

src/lsp4clj/server.clj
Cyrik: I could probably look all these up myself, but I think you have a better overview, so here goes:
1. How difficult is it to figure out if we're dropping messages?
2. Could we select which messages to drop, e.g. drop stuff like `codeLense` to keep space for `didChange`?
3. Should we make this configurable? I'm guessing a server-based approach could handle a different size than a local one.

That being said, thinking about this is probably overkill and @ericdallo is most likely right that we'll never hit the size.

src/lsp4clj/server.clj
Cyrik: Ah, I should have read your message before writing mine.
I agree that it would be ideal if we could error if the buffer is full instead of dropping the message.

The new solution is already much better than the previous one, so it's not urgent, but we should get to it at some point.

ericdallo: Agreed, throwing an exception in that case seems better indeed

src/lsp4clj/server.clj
mainej: @ferdinand-beyer, so that I'm clear, this is a problem for Scarlet, but not for clojure-lsp, right? The difference being that clojure-lsp always shuts down the whole process after this cleanup is done, which ends up killing threads blocked waiting for a response. In contrast, Scarlet has a long-running server that manages the lifecycle of many lsp4clj servers. It shuts them down and restarts others without itself stopping entirely. So, it has to be more careful about thread management. Sound accurate?

I think you're worried, in particular, about threads waiting to `deref-or-cancel` or `deref` a pending-sent-request... is that correct? If so, it may be possible to `future-cancel` all those pending-sent-requests. That would `(deliver p ::cancelled)` to their promises, thus unblocking their waiters. If we do that, we might drop actual buffered client responses, a problem which the "poison pill" might avoid. But on the other hand, cancelling seems more direct, and since we're shutting down, the last few responses are probably irrelevant. (Basically, I'm not sure I could figure out how to implement a poison pill, but cancelling would be straightforward. 😅 )

mainej: (And, FWIW, the lsp4clj docs are pretty clear that you should prefer `deref-or-cancel`, which means that blocked threads wouldn't stay blocked forever.)

src/lsp4clj/server.clj
mainej: > Have multiple readers for client-initiated-ch

As everyone prefers the second solution, we shouldn't dwell on this one. But for the record, it would cause problems, I think. If you have multiple threads reading from the `client-initiated-ch`, you could receive and start processing a refactoring via `executeCommand` before finishing a `didChange`. You'd end up refactoring an old version of the document. lsp4clj needs to enforce sequential processing sometimes, and having a single reader from the `client-initiated-ch` is an important part of that.

> Detect that buffers are full and raise an error, so that any server threads waiting for responses get unblocked. This is probably a bit tricky to implement, potentially using offer! (or alts!/alt!) and replacing the promise in PendingRequest with a promesa deferred

I agree, dropping messages isn't great. I thought about that a little as I was implementing this, but got overwhelmed considering all the possibilities. I think it'd be good to collaborate on a solution, because I'm not sure I know what to do.

I kind of see what you mean by raising exceptions, but thinking about it more carefully, we need to clarify—where should the exception be raised? In what thread(s)?

I just took a very quick pass at detecting that the buffers are full. As you suggested, we can do it by using [`offer!` instead of `>!`](https://github.com/clojure-lsp/lsp4clj/blob/ac2deb76ded6d3d0f8fced3643bfe7836152a4be/src/lsp4clj/server.clj#L183-L185) on the client- and server-initiated channels. But then what?

* We're in the [pipeline thread](https://github.com/clojure-lsp/lsp4clj/blob/ac2deb76ded6d3d0f8fced3643bfe7836152a4be/src/lsp4clj/server.clj#L176). We can stop that thread by returning instead of recurring. (Taking from an `async/thread` that has thrown an exception just returns nil, so when you detect a full buffer, throwing an exception is the same as returning nil.) From there we could do all the [cleanup](https://github.com/clojure-lsp/lsp4clj/blob/ac2deb76ded6d3d0f8fced3643bfe7836152a4be/src/lsp4clj/server.clj#L267-L293) in `start`. Then we could [deliver](https://github.com/clojure-lsp/lsp4clj/blob/ac2deb76ded6d3d0f8fced3643bfe7836152a4be/src/lsp4clj/server.clj#L294-L297) `:buffer-full` to the promise returned by `start`. But we don't have any guarantee that anyone is holding onto, or deref-ing that return value. The lsp4clj server would stop doing anything, but without cooperation from Scarlet or clojure-lsp no exception would be raised, and no error would get logged. The language server wouldn't necessarily stop running.
* Alternatively, you could log an error and do all the same stuff. 
* Alternatively, you could try to figure out how to raise exceptions elsewhere in the language server, in a way that would cause everything to wrap up. I'm not quite sure how that would work or where the exceptions would occur—perhaps another variant of the poison pill idea?

Anyway, again, I'd appreciate some help thinking through exactly what should happen.

Cyrik: IMHO, it'd be nice to have a very basic solution quickly and see if this is actually a problem. Otherwise we either wait a long time to have something or we burn a bunch of time on something that barely ever happens.
So my vote would be to log and explode as much as is easily achievable.
If we see those errors actually occurring we can start spending real effort.

ferdinand-beyer: I'd go with trying the following (I can also draft a PR if this helps):

- Change the `promise` in `PendingRequest` to a `promesa.core/deferred`. Now it can be delivered with `p/resolve!` or it can be put to an error state with `p/reject!`
- Write a function `fail-pending-sent-requests` that rejects all promises in the `pending-sent-request*` atom
- Change the buffer type to a normal fixed-size buffer
- Replace the `>!` in the pipeline with:

  ```clojure
  (when-not (async/offer! x-initiated-ch the-msg)
    ;; Buffers full. Fail any waiting pending requests and block.
    (fail-pending-sent-requests (ex-info "response not received in time" {}))
    ;; Try again, but block this time, to apply back pressure to our client
    (async/>! x-initiated-ch the-msg))
  ```

This will trigger the exception in any handler threads that are waiting for a response that we could not deliver. In handler code, this would be very similar to a timeout. I would not fail the pipeline, but as shown fall back to blocking, so that we can apply back pressure.

The function `fail-pending-sent-requests` could be reused in the clean-up phase, to interrupt any waiting threads.

src/lsp4clj/server.clj
ferdinand-beyer: > this is a problem for Scarlet, but not for clojure-lsp, right?

Right now, yes. But I don't know who else is using lsp4clj, so there might be others in a similar situation. `deref-or-cancel` should indeed solve this, running into a timeout eventually.

I think `future-cancel` would cause more harm than good, as clients certainly don't expect to get `::cancelled` as a "magic" response value. If they use the 3-arity `deref` or `deref-or-cancel`, they would get their expected timeout value, not `::cancelled`.

The "poison pill" approach could indeed get ugly, as messages remaining in the buffer could cause a handler to send another request and wait forever for the response, with the poison stuck in the buffer. It's definitely easier to just fail pending requests right away, using the `fail-pending-sent-requests` function I suggested in the other thread.

src/lsp4clj/server.clj
ferdinand-beyer: A nice side-effect of using a promesa promise inside `PendingRequest`: We could change it to be a promise itself.

We can piggy-back on the underlying `CompletableFuture`, replacing the `cancelled?` atom with `p/cancel!` and `p/cancelled?`, and implementing `IPromiseFactory` ourselves, returning the underlying promise.

Now users can use promesa to chain callbacks to the request, eliminating the need for blocking in some situations.

PR #38, opened 2023-03-27
Avoid dropping client messages

mainej:
@ferdinand-beyer this is a draft of the changes you [suggested](https://github.com/clojure-lsp/lsp4clj/pull/37#discussion_r1136615400). Is it roughly what you were thinking? I have some questions/concerns below about whether all of this is the right way to go. Will you review?

## Pending Sent Requesets

* A pending sent request is now mostly a wrapper around a promesa deferred, and can be used as one.
* That simplified many things, but I couldn't figure out how to make `p/cancel!` send `$/cancelRequest`. It'd be nice (and intuitive) if it did. 
  * `$/cancelRequest` is still exposed through `deref-or-cancel`, which, under the hood, uses `java.util.concurrent.Future/cancel`, or equivalently, `clojure.core/future-cancel`.
  * I wasn't sure how to make the pending sent request both _return_ a promise via `IPromiseFactory` and _be_ a promise in the sense of implementing `promesa.protocols/ICancellable`. If I were to do it again, I don't think I'd make `PendingRequest` implement so many interfaces, because promesa's promises already implement many of those same interfaces. Instead, I think I'd prefer if you could get to those interfaces via the promise. Should we just move forward with that? It might be best to do in a v2.
* If the client returns an error, the pending request is resolved (not rejected as you might expect.) That's because:
  * `p/reject!` expects an exception, whereas the client's response isn't exactly an exception, but is instead a result with a special interpretation—that an error of some sort happened on the client side,
  * and, because the language servers are designed to `deref` an error _result_, not `(p/catch ...)` an _exception_.

## Dropping messages

* As you suggested, when the buffer of client-initiated input fills up we now reject our pending sent requests, then park.
* However, I'm having trouble reasoning about whether this strategy really makes sense. We suspect that the reason the client-initiated input has filed up is because the server is blocking on something, and not getting to the client's messages. But a pending sent request is only one of many things the server might be blocking on. Cancelling all the pending sent requests _might_ unblock the server, but the server might have been blocked on something else entirely. In fact, there might not even be any pending sent requests to reject.
* Your original proposal treated the server-initiated input symmetrically, rejecting the client's messages.
  * On reflection, I don't think that's advisable. That would prioritize the server's messages over the client's, which isn't really what we want.
  * So, I've left that part as a sliding-buffer. However, perhaps it should be a fixed-size buffer. In either case, I'm still not sure how to reason about how big that buffer should be.
* Originally, the sliding-buffer was 100 messages. I bumped this number to 1024, and I'm using it for both the sliding-buffer for server-initiated messages, and as the fixed-size buffer for client-initiated messages. I have no reason for this change, except that it's the number that core.async uses internally for when it starts rejecting `put!`s.

This whole series of changes makes me wonder if lsp4clj would be better if it were built mostly on promesa (or manifold, or...), not on core.async. It feels like we have an unholy mixture of the two at the moment. Switching those underpinnings might not even involve many changes to `clojure-lsp` and Scarlet, because they don't really work down at the `chan-server` level, which is the current base of the lsp4clj functionality. On the other hand, it might just be trading one set of problems for another. Something to think about for v2, if that day ever comes.

### DIFF
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -90,10 +90,10 @@
   ;; Chaining `(-> (p/deferred) (p/catch ...))` seems like it should work, but
   ;; doesn't.
   (let [p (p/deferred)]
+    ;; side-effect of cancellation:
     (p/catch p CancellationException
-      (fn [ex]
-        (protocols.endpoint/send-notification server "$/cancelRequest" {:id id})
-        (p/rejected ex)))
+      (fn [_]
+        (protocols.endpoint/send-notification server "$/cancelRequest" {:id id})))
     (map->PendingRequest {:p p
                           :id id
                           :method method
@@ -221,11 +221,9 @@
           ;; If the server blocks waiting for a response, we have to set aside
           ;; the other inbound requests and notifications, so that we can get to
           ;; the response. That is, while the server is blocking we cannot stop
-          ;; reading input. Otherwise, the server will end up in a deadlock,
-          ;; where it's waiting to receive a response off the input-ch but new
-          ;; messages aren't being put on the input-ch because the server is
-          ;; blocked. See
-          ;; https://github.com/clojure-lsp/clojure-lsp/issues/1500.
+          ;; accepting input. Otherwise, the server will end up in a deadlock,
+          ;; where it's waiting to receive a response, but the response is
+          ;; waiting to be accepted.
 
           ;; To accomplish this we processes inbound requests and notifications
           ;; separately from inbound responses. If the server starts blocking
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -9,7 +9,7 @@
    [lsp4clj.protocols.endpoint :as protocols.endpoint]
    [lsp4clj.trace :as trace]
    [promesa.core :as p]
-   [promesa.protocols])
+   [promesa.protocols :as p.protocols])
   (:import
    (java.util.concurrent CancellationException)))
 
@@ -31,7 +31,7 @@
 (defprotocol IBlockingDerefOrCancel
   (deref-or-cancel [this timeout-ms timeout-val]))
 
-(defrecord PendingRequest [p cancelled? id method started server]
+(defrecord PendingRequest [p id method started server]
   clojure.lang.IDeref
   (deref [_] (deref p))
   clojure.lang.IBlockingDeref
@@ -39,36 +39,31 @@
     (deref p timeout-ms timeout-val))
   IBlockingDerefOrCancel
   (deref-or-cancel [this timeout-ms timeout-val]
-    (let [result (deref this timeout-ms ::timeout)]
-      (if (= ::timeout result)
+    (let [result (deref p timeout-ms ::timeout)]
+      (if (identical? ::timeout result)
         (do (future-cancel this)
             timeout-val)
         result)))
   clojure.lang.IPending
-  (isRealized [_] (realized? p))
+  (isRealized [_] (p/done? p))
   java.util.concurrent.Future
-  (get [this]
-    (let [result (deref this)]
-      (if (= ::cancelled result)
-        (throw (java.util.concurrent.CancellationException.))
+  (get [_] (deref p))
+  (get [_ timeout unit]
+    (let [result (deref p (.toMillis unit timeout) ::timeout)]
+      (if (identical? ::timeout result)
+        (throw (java.util.concurrent.TimeoutException.))
         result)))
-  (get [this timeout unit]
-    (let [result (deref this (.toMillis unit timeout) ::timeout)]
-      (case result
-        ::cancelled (throw (java.util.concurrent.CancellationException.))
-        ::timeout (throw (java.util.concurrent.TimeoutException.))
-        result)))
-  (isCancelled [_] @cancelled?)
-  (isDone [this] (or (.isRealized this) (.isCancelled this)))
-  (cancel [this _interrupt?]
-    (if (.isDone this)
+  (isCancelled [_] (p/cancelled? p))
+  (isDone [_] (p/done? p))
+  (cancel [_ _interrupt?]
+    (if (p/done? p)
       false
-      (if (compare-and-set! cancelled? false true)
-        (do
-          (protocols.endpoint/send-notification server "$/cancelRequest" {:id id})
-          (deliver p ::cancelled)
-          true)
-        false))))
+      (do
+        (p/cancel! p)
+        (protocols.endpoint/send-notification server "$/cancelRequest" {:id id})
+        true)))
+  p.protocols/IPromiseFactory
+  (-promise [_] p))
 
 ;; Avoid error: java.lang.IllegalArgumentException: Multiple methods in multimethod 'simple-dispatch' match dispatch value: class lsp4clj.server.PendingRequest -> interface clojure.lang.IPersistentMap and interface clojure.lang.IDeref, and neither is preferred
 ;; Only when CIDER is running? See https://github.com/thi-ng/color/issues/10
@@ -96,8 +91,7 @@
   Sends `$/cancelRequest` only once, though `lsp4clj.server/deref-or-cancel` or
   `future-cancel` can be called multiple times."
   [id method started server]
-  (map->PendingRequest {:p (promise)
-                        :cancelled? (atom false)
+  (map->PendingRequest {:p (p/deferred)
                         :id id
                         :method method
                         :started started
@@ -222,7 +216,7 @@
     (async/put! trace-ch [:debug trace-body])))
 
 (defrecord PendingReceivedRequest [result-promise cancelled?]
-  promesa.protocols/ICancellable
+  p.protocols/ICancellable
   (-cancel! [_]
     (p/cancel! result-promise)
     (reset! cancelled? true))
@@ -330,7 +324,7 @@
         (if-let [{:keys [p started] :as req} (get pending-requests id)]
           (do
             (trace this trace/received-response req resp started now)
-            (deliver p (if error resp result)))
+            (p/resolve! p (if error resp result)))
           (trace this trace/received-unmatched-response resp now)))
       (catch Throwable e
         (log-error-receiving this e resp))))
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -38,10 +38,10 @@
   (deref [_ timeout-ms timeout-val]
     (deref p timeout-ms timeout-val))
   IBlockingDerefOrCancel
-  (deref-or-cancel [this timeout-ms timeout-val]
+  (deref-or-cancel [_ timeout-ms timeout-val]
     (let [result (deref p timeout-ms ::timeout)]
       (if (identical? ::timeout result)
-        (do (future-cancel this)
+        (do (p/cancel! p)
             timeout-val)
         result)))
   clojure.lang.IPending
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -228,7 +228,7 @@
           ;; To accomplish this we processes inbound requests and notifications
           ;; separately from inbound responses. If the server starts blocking
           ;; waiting for a response, we buffer the inbound requests and
-          ;; notificatons until the server is prepared to process them.
+          ;; notifications until the server is prepared to process them.
 
           ;; If the buffer becomes full, we assume that the server isn't
           ;; handling inbound requests and notifcations because it's waiting for
@@ -241,7 +241,7 @@
 
           ;; This ensures we don't drop any client messages, though we could
           ;; stop reading them if the server keeps blocking. If we're lucky
-          ;; either the language server will unblock, or the client will decided
+          ;; either the language server will unblock, or the client will decide
           ;; to stop sending messages because it's failed to receive a server
           ;; response (i.e., we will have managed to apply backpressure to the
           ;; client). If we're unlucky, the server could keep blocking forever.
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -199,56 +199,57 @@
   protocols.endpoint/IEndpoint
   (start [this context]
     ;; Start receiving messages.
-    (let [;; In order to process some requests and (all) notifications in
-          ;; series, the language server sometimes needs to block
-          ;; client-initiated input. If the language server sends requests
-          ;; during that time, it needs to receive responses, even though it's
-          ;; blocking other input. Otherwise, it will end up in a deadlock,
-          ;; where it's waiting to receive a response off the input-ch and the
-          ;; input-ch isn't being read from because the server is blocking
-          ;; input. See https://github.com/clojure-lsp/clojure-lsp/issues/1500.
+    (let [;; The language server sometimes needs to stop processing inbound
+          ;; requests and notifications.
 
-          ;; To avoid this problem we processes client-initiated input (client
-          ;; requests and notifications) and server-initiated input (client
-          ;; responses) separately.
+          ;; We do this automatically whenever we receive a notification, so
+          ;; that we are sure the language server processes didChange before
+          ;; moving on to other requests. But the language server can also
+          ;; decide to do it itself, even in a request, by processing everything
+          ;; synchronously instead of returning a future.
 
-          ;; If the server starts blocking waiting for a response, we buffer the
-          ;; client's requests and notifications until the server is prepared to
-          ;; process them.
+          ;; In a request (or even in a notification) the language server can
+          ;; send a request to the client and block waiting for a response.
 
-          ;; However, if too many client requests and notifications arrive
-          ;; before the response, the buffer fills up. In this situation, we
-          ;; abort all the server's pending requests, hoping that this will
-          ;; allow it to process the client's other messages. We do this to
-          ;; prioritize the client's messages over the server's.
+          ;; It's possible -- even likely -- that the client will send other
+          ;; messages between the time that the server sent its request and the
+          ;; time that the client responds.
 
-          ;; The situation is different in reverse. The client can also start
-          ;; blocking waiting for a server response. In the meantime, the server
-          ;; can send lots of messages. But this is bad behavior on the server's
-          ;; part. So in this scenario we drop the server's earliest requests
-          ;; and notifications.
+          ;; All inbound messages -- requests, notifications, and responses --
+          ;; come in on a single stream.
 
-          ;; We do that by storing them in a sliding buffer. Because of the
-          ;; sliding buffer:
-          ;; * if the language server sends a message, and
-          ;; * if while processing that message the client sends a request which
-          ;;   causes it to block, and
-          ;; * if the language server sends too many other messages between when
-          ;;   the client started blocking and when the language server responds
-          ;;   to the client's request,
-          ;; * then the language server's messages will start being dropped,
-          ;;   starting from the earliest.
+          ;; If the server blocks waiting for a response, we have to set aside
+          ;; the other inbound requests and notifications, so that we can get to
+          ;; the response. That is, while the server is blocking we cannot stop
+          ;; reading input. Otherwise, the server will end up in a deadlock,
+          ;; where it's waiting to receive a response off the input-ch but new
+          ;; messages aren't being put on the input-ch because the server is
+          ;; blocked. See
+          ;; https://github.com/clojure-lsp/clojure-lsp/issues/1500.
 
-          ;; We process the client- and language-server-initiated messages in
-          ;; separate threads.
-          ;; * Threads, so the language server can use >!! and so that we can use
-          ;;   (>!! output-ch) to respect back pressure from clients that are slow
-          ;;   to read.
-          ;; * Separate, so one can continue while the other is blocked.
-          server-initiated-in-ch (thread-loop
-                                   (async/sliding-buffer input-buffer-size)
-                                   (fn [response]
-                                     (protocols.endpoint/receive-response this response)))
+          ;; To accomplish this we processes inbound requests and notifications
+          ;; separately from inbound responses. If the server starts blocking
+          ;; waiting for a response, we buffer the inbound requests and
+          ;; notificatons until the server is prepared to process them.
+
+          ;; If the buffer becomes full, we assume that the server isn't
+          ;; handling inbound requests and notifcations because it's waiting for
+          ;; a response. So, following our assumption, we reject the server
+          ;; requests so that the server will stop waiting for the response.
+          ;; It's possible that this won't work -- our assumption might have
+          ;; been wrong and the server might have stalled for some other reason.
+          ;; So after rejecting, we park trying to add the latest inbound
+          ;; request or notification to the buffer.
+
+          ;; This ensures we don't drop any client messages, though we could
+          ;; stop reading them if the server keeps blocking. If we're lucky
+          ;; either the language server will unblock, or the client will decided
+          ;; to stop sending messages because it's failed to receive a server
+          ;; response (i.e., we will have managed to apply backpressure to the
+          ;; client). If we're unlucky, the server could keep blocking forever.
+          ;; In any case, this scenario -- where we stop reading messages -- is
+          ;; presumed to be both very rare and indicative of a problem that can
+          ;; be solved only in the client or the language server.
           client-initiated-in-ch (thread-loop
                                    input-buffer-size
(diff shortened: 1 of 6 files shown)


ferdinand-beyer:
> Is it roughly what you were thinking?

Absolutely! Thank you for considering my suggestions!

> That simplified many things, but I couldn't figure out how to make `p/cancel!` send `$/cancelRequest`. It'd be nice (and intuitive) if it did.

I think we can make use of the fact that on the JVM, `p/deferred` creates a `CompletableFuture`, and `p/cancel!` just calls `CompletableFuture.cancel()`. This in turn just makes the promise error with a `java.util.concurrent.CancellationException`, if it is not completed yet.

We can therefore detect if a promise was cancelled by catching this exception, and send `$/cancelRequest`:

```clojure
(map->PendingRequest {:p (-> (p/deferred)
                             (p/catch java.util.concurrent.CancellationException
                                      (fn [ex]
                                        (protocols.endpoint/send-notification server "$/cancelRequest" {:id id})
                                        (p/rejected ex))))
                      :id id
                      :method method
                      :started started
                      :server server}))
```

Our own implementation of `cancel` should then just delegate to `p/cancel!`.

I think this is the right thing to do anyway, as we have a race condition when cancelling, and probably don't want to send the notification when the request is already completed between `(p/done?)` and `(p/cancel!)`.

> I wasn't sure how to make the pending sent request both return a promise via `IPromiseFactory` and be a promise in the sense of implementing `promesa.protocols/ICancellable`. If I were to do it again, I don't think I'd make `PendingRequest` implement so many interfaces, because promesa's promises already implement many of those same interfaces. Instead, I think I'd prefer if you could get to those interfaces via the promise. Should we just move forward with that? It might be best to do in a v2.

I agree. Implementing `IPromiseFactory` should be enough for many use cases, as you can coerce it to a promesa promise and then use its functions.

I am wondering if we could actually get rid of `PendingRequest` altogether, and just use a plain promesa promise. If I'm not mistaken, this should already implement most protocols and interfaces that `PendingRequest` does, with the exceptions being `IPending` and `IBlockingDerefOrCancel`:

- `IDeref` is not directly implemented, but `clojure.core/deref` will work on futures
- `IPending` -- too bad this is an interface and not a protocol, so it seems we cannot support `realized?` without a custom type.
- `IBlockingDerefOrCancel` could be implemented for `java.util.concurrent.Future`, thus working for all futures including Promesa promises. It would not even need to be a protocol, but could be a plain function.

> If the client returns an error, the pending request is resolved (not rejected as you might expect.)

That's indeed surprising. I think it would be cleaner to reject with an `ex-info` and the client result in `ex-data`. However, as you pointed out, this would probably break existing code.

ferdinand-beyer:
> However, I'm having trouble reasoning about whether this strategy really makes sense. We suspect that the reason the client-initiated input has filed up is because the server is blocking on something, and not getting to the client's messages. But a pending sent request is only one of many things the server might be blocking on. Cancelling all the pending sent requests might unblock the server, but the server might have been blocked on something else entirely. In fact, there might not even be any pending sent requests to reject.

That's right, we don't know at this point why the server was not fast enough to process all messages.

I think that ideally the message pipeline should not guess and just trust that the server will unblock eventually. The only exception is if the server would be blocked because of the pipeline's implementation, and failing pending client responses _could_ solve that. With "solving" I actually mean "help diagnose" because I think this situation should never happen.

> So, I've left that part as a sliding-buffer.

Not sure if I follow your reasoning, I will comment on the code changes.

src/lsp4clj/server.clj
```diff
@@ -259,14 +197,95 @@
...
+
+          ;; If the server starts blocking waiting for a response, we buffer the
+          ;; client's requests and notifications until the server is prepared to
+          ;; process them.
+
+          ;; However, if too many client requests and notifications arrive
+          ;; before the response, the buffer fills up. In this situation, we
+          ;; abort all the server's pending requests, hoping that this will
+          ;; allow it to process the client's other messages. We do this to
+          ;; prioritize the client's messages over the server's.
```
ferdinand-beyer: Do you mean we prioritize because we abort pending server requests, but never abort pending client requests?

I think it is worth mentioning that we only do that because we assume the server could be blocked waiting for a client response, and we somehow need to keep the input channel open or else we would make it impossible for us to receive the response.

src/lsp4clj/server.clj
```diff
@@ -259,14 +197,95 @@
...
+
+          ;; The situation is different in reverse. The client can also start
+          ;; blocking waiting for a server response. In the meantime, the server
```
ferdinand-beyer: True that the client could block, but from the server's perspective I think it is the client's responsibility to handle that situation. Either it never blocks (VS-Code/JavaScript), or could use some dropping buffer of its own.

src/lsp4clj/server.clj
```diff
@@ -259,14 +197,95 @@
...
+          ;; We do that by storing them in a sliding buffer. Because of the
+          ;; sliding buffer:
+          ;; * if the language server sends a message, and
+          ;; * if while processing that message the client sends a request which
+          ;;   causes it to block, and
+          ;; * if the language server sends too many other messages between when
+          ;;   the client started blocking and when the language server responds
+          ;;   to the client's request,
+          ;; * then the language server's messages will start being dropped,
+          ;;   starting from the earliest.
```
ferdinand-beyer: I think this is incorrect.

What you describe sounds like a sliding buffer for messages _sent by the server_, but here we are using a sliding buffer for messages _sent by the client_. The sliding buffer causes the server to drop client responses.

If I understand you correctly you are describing a scenario where the client waits on the server, therefore stops processing messages from the server, causing back pressure on the server, which manifests as the server getting blocked on its output channel. I don't think this can be resolved by the server.

It seems to me that a LSP _client_ would be ill advised to block indefinitely on a server response, as the client provides the user experience and could decide to ignore old server messages.

I think the lsp4clj pipeline should never silently drop messages.

src/lsp4clj/server.clj
```diff
@@ -330,7 +349,7 @@
         (if-let [{:keys [p started] :as req} (get pending-requests id)]
           (do
             (trace this trace/received-response req resp started now)
-            (deliver p (if error resp result)))
+            (p/resolve! p (if error resp result)))
```
ferdinand-beyer: If I understand you correctly we do not `reject!` on error because the former implementation did not support rejected promises, and we don't want to break existing code.

Should we document this here?

Maybe we could add an option to `send-request` to throw exceptions on error, similar to how `clj-http` supports `:throw-exceptions? false`. Then new code could set this to `true` and eventually we could make this the default?

mainej:
> `(-> (p/deferred) (p/catch java.util.concurrent.CancellationException ...))`

Thanks for the tip. My intuition was that your suggestion—setting `p` to `(-> (p/deferred) (p/catch ...)`—should work. But that catch is never invoked, even when `p` is cancelled. I found a workaround, which you can now see in the code. You have to set `p` to `(p/deferred)` and separately call `(p/catch p ...)`. Weird.

> I am wondering if we could actually get rid of `PendingRequest` altogether, and just use a plain promesa promise

Perhaps, although I'm not sure. The `PendingRequest` is a wrapper for the promise, but also for `:method`, `:id`, and `:start`, which are used in tracing. But maybe there's another way to structure things so that we can just use a plain promise. It's something to experiment with in a v2.

> reject with an `ex-info` and the client result in `ex-data`.

Agreed... another thing to consider in a v2.

Thanks for your other notes. I don't have time to consider them carefully right now, but will as soon as I can.

src/lsp4clj/server.clj
mainej: I'll remove the "prioritize" language, and clarify this comment as you suggested.

src/lsp4clj/server.clj
mainej: Yeah, apologies, this comment is incorrect, and makes it much harder to understand what's going on.

The situation at the moment is that inbound requests and notifications are added to a fixed size buffer. If the buffer becomes full, we assume that the server isn't handling inbound requests and notifications because it's waiting for an inbound response. So, following our assumption, we reject the outbound requests so that the server will stop waiting for the inbound response. It's possible that this won't work—our assumption might have been wrong and the server might have stalled for some other reason. So after rejecting, we park trying to add the latest inbound request or notification to the buffer.

Inbound responses are a different story. At the moment, they're added to a sliding buffer. If the buffer becomes full, we start dropping inbound responses.

@ferdinand-beyer I think, though I'd like clarification, that you're saying you don't think we should drop inbound responses. That makes sense to me.

The question becomes, what _should_ we do? After thinking about this a bit, I don't think it's even possible for that buffer to fill up. We immediately take client responses off that channel and give them to `protocols.endpoint/receive-response`. But that never blocks, and in particular, doesn't call anything in the language server that might block. 

So, I think we can just call `protocols.endpoint/receive-response` directly, bypassing any intermediary buffers. (That is, we can remove `server-initiated-input-ch`.) See anything wrong with that?

src/lsp4clj/server.clj
mainej: > we do not reject! on error because the former implementation did not support rejected promises, and we don't want to break existing code

Correct, existing code expects to `(let [{:keys [error] :as resp} (deref-or-cancel req ...`.

> Maybe we could add an option to `send-request` to throw exceptions on error

You mean reject on error, so that dereffing would throw? That might be nice, although I don't want to add it in this PR. Something like that could be a per-request setting, as part of `send-request`, or it could be a global setting, as part of `chan-server`.

> Should we document this here?

Yes, I'll at least add some documentation.

src/lsp4clj/server.clj
mainej: **Edit**: I've removed `server-initiated-input-ch`. But still please let me know if you see a reason we'd need it.

src/lsp4clj/server.clj
mainej: **Edit:** comments added.

src/lsp4clj/server.clj
mainej: I've updated this comment. Hopefully it's clearer now?

ferdinand-beyer:
> My intuition was that your suggestion—setting `p` to `(-> (p/deferred) (p/catch ...)`—should work. But that catch is never invoked, even when `p` is cancelled. I found a workaround, which you can now see in the code. You have to set `p` to `(p/deferred)` and separately call `(p/catch p ...)`. Weird.

Now that you say it, it actually makes sense. I did not think about that.

We need to remember that promises form a chain. Let's call the promise returned by `(p/deferred)` `p0` and the one returned by `(p/catch CancellationException ...)` `p1`. Now `p1` will handle the cancellation of `p0`. But for that to work, we need to actually cancel `p0`, not `p1`, and therefore we need to keep `p0` around.

When we chain and cancel `p1`, its own callback will never be called.

We can actually "fork" `p0` for user code and ignore the value that `p1` yields, as it will only be used for side effects.

src/lsp4clj/server.clj
```diff
@@ -31,44 +31,35 @@
...
-        (do
-          (protocols.endpoint/send-notification server "$/cancelRequest" {:id id})
-          (deliver p ::cancelled)
-          true)
-        false))))
+  (isCancelled [_] (p/cancelled? p))
+  (isDone [_] (p/done? p))
+  (cancel [_ _interrupt?]
+    (p/cancel! p)
+    (p/cancelled? p))
```
ferdinand-beyer, line 63: I think `p/cancelled?` is unnecessary, as `p/cancel!` will already return the return value of the underlying `CompleteableFuture`, which itself implements `Future`. We can just do `(cancel [_ _] (p/cancel! p))`

src/lsp4clj/server.clj
```diff
@@ -96,12 +87,17 @@
...
-                        :method method
-                        :started started
-                        :server server}))
+  ;; Chaining `(-> (p/deferred) (p/catch ...))` seems like it should work, but
+  ;; doesn't.
+  (let [p (p/deferred)]
+    (p/catch p CancellationException
+      (fn [ex]
+        (protocols.endpoint/send-notification server "$/cancelRequest" {:id id})
+        (p/rejected ex)))
```
ferdinand-beyer: We "fork" `p` to handle cancellation. The promise returned by `p/catch` will never be used. We can therefore return whatever, and can remove the `(p/rejected ex)` form.

src/lsp4clj/server.clj
```diff
@@ -259,13 +199,91 @@
...
+          ;; All inbound messages -- requests, notifications, and responses --
+          ;; come in on a single stream.
+
+          ;; If the server blocks waiting for a response, we have to set aside
+          ;; the other inbound requests and notifications, so that we can get to
+          ;; the response. That is, while the server is blocking we cannot stop
+          ;; reading input. Otherwise, the server will end up in a deadlock,
+          ;; where it's waiting to receive a response off the input-ch but new
+          ;; messages aren't being put on the input-ch because the server is
+          ;; blocked. See
```
ferdinand-beyer: I think this is not quite accurate. Messages keep being put on `input-ch`, but they are queued 

> where' it's waiting to receive a _response_ off the input-ch  **that is blocked by other messages before it in the queue**

src/lsp4clj/server.clj
ferdinand-beyer: Unfortunately promesa, like similar libraries, is not very good at documenting return values of these operations.

src/lsp4clj/server.clj
ferdinand-beyer: > I think, though I'd like clarification, that you're saying you don't think we should drop inbound responses.

Yes, this is what I meant 👍 

> So, I think we can just call `protocols.endpoint/receive-response` directly, bypassing any intermediary buffers. (That is, we can remove `server-initiated-input-ch`.) See anything wrong with that?

Looks fine to me! The only thing that comes to mind is that we rely a lot on known implementations of a _protocol_. If we ever use a different implementation, or change the way `receive-notification` is implemented, we need to keep this in mind. Maybe in a v2 we should reconsider using protocols here if we don't expect multiple implementations.

src/lsp4clj/server.clj
ferdinand-beyer: Thanks! Makes sense!

I'm wondering though, if we should move this detail to the `deref` implementations of `PendingRequest`, and `p/reject!` here on error. This way, new code could use the promesa interface of `PendingRequest` and handle errors as expected, while old code, before `PendingRequest` implemented `IPromiseFactory`, would work as before.

In other words, `(deref pending-request)` would return a map with an `:error`, while `(deref (p/promise pending-request))` would throw.

I think that I, personally, would treat `PendingRequest` as a promesa promise after this PR is merged, and stop using the old interface.

src/lsp4clj/server.clj
mainej: > we rely a lot on known implementations of a `protocol`

I'm not sure that's quite right... this code knows how _this_ implementation of the protocol works, but other implementors of the protocol aren't expected to use this code.

We could remove the protocol. It isn't buying us much. That said, `clojure-lsp` has a test helper which implements it. (That implementation acts as a fake client in integration tests. The protocol is designed to represent either side of LSP comms, a client sending and receiving messages from a server, but also a server sending and receiving messages from a client.) And so in a sense the protocol acts as a tiny amount of documentation that both tools, the server and the fake client, have the same responsibilities. It's not a huge benefit, but doesn't hurt either.

src/lsp4clj/server.clj
mainej: I'm going to leave this as is for now. It'd be good to revisit in a v2, or I'd be happy to review a PR.

> I, personally, would treat PendingRequest as a promesa promise after this PR is merged, and stop using the old interface.

As soon as you convert the PendingRequest to a promesa promise, you lose access to `deref-or-cancel`. To replicate that, you'd have a few options.
* You could write your own code that added a timeout to the promise, caught the timeout, and did a cancellation.
* You could add more code to lsp4clj that encapsulated that pattern, although you'd basically be back to where we are now.

src/lsp4clj/server.clj
mainej: I don't think that's equivalent. `p/cancel!` returns the CompletableFuture object, but the Future interface says `.cancel` returns a boolean.

src/lsp4clj/server.clj
mainej: I think you might be confusing the `input-ch` and the new `client-initiated-in-ch`. The `input-ch` is a `(chan 1)`, fed by a `>!!` in `io-chan/input-stream->input-chan`, so it doesn't have room to queue anything. The new `client-initiated-in-ch` has a buffer and can queue messages. If you block between reading off the `input-ch`, as we were doing before, the feeder literally cannot put anything else onto it, not even into a queue.

That said, this comment is a bit hard to understand, as it's talking about a scenario that we used to have but no longer do. I'll edit it to talk more about the problem we're avoiding, and less about how specific configurations of queues can cause it.

mainej:
> We need to remember that promises form a chain.

That makes sense. Thanks for the explanation!

src/lsp4clj/server.clj
ferdinand-beyer: Ah, you are right! I might have overlooked that `p/cancel` calls the implementation, ignores its return value, and returns the promise. Or maybe I looked at a different version of promesa...

src/lsp4clj/server.clj
ferdinand-beyer: > I think you might be confusing the `input-ch` and the new `client-initiated-in-ch`. The `input-ch` is a `(chan 1)`, fed by a `>!!` in `io-chan/input-stream->input-chan`, so it doesn't have room to queue anything.

Users can use different implementations that `io-chan` (e.g. websockets), and could use different buffer strategies for the `input-ch`.

Anyway, thanks for improving this!

src/lsp4clj/server.clj
ferdinand-beyer: I think implementing `deref-or-cancel` for `CompletableFuture` is trivial, and could be provided by `lsp4clj`. This could be a step towards getting rid of `PendingRequest` altogether.

To be honest, but this might be a different topic, I'm not convinced that `deref-or-cancel` should be used anyway. When we do not get a response in time, chances are high that there was a problem with the original request, and I'm not sure if the client would actually benefit from knowing that the server is no longer interested in the response. Do we have evidence from clojure-lsp that this is helpful?

Sending a cancel actually adds congestion, as we will block on the output channel to send the cancel notification. We could just as well go on and ignore any response that comes in later.

mainej: > I think implementing deref-or-cancel for CompletableFuture is trivial... This could be a step towards getting rid of PendingRequest altogether.

Yeah, sounds like a nice addition. I've run out of steam in this PR though. :)

> We could just as well go on and ignore any response that comes in later.

We do ignore the response, whether we timeout using a `deref` with a timeout, or with `deref-or-cancel`.

> I'm not sure if the client would actually benefit from knowing that the server is no longer interested in the response. Do we have evidence from clojure-lsp that this is helpful?

The server doesn't benefit from sending the cancellation. It's a courtesy to the client, potentially allowing it to skip or abort some work. I agree, it's unclear whether the client really benefits from that though. It's possible things are so broken on the client that a single cancellation won't help it. On the other hand, lsp4clj is a low-level tool, and so it seems like it should stay close to the spec. The spec suggests that the server should _try_ to send a cancellation, so I think it's a facility lsp4clj should offer. Of course higher level tools like clojure-lsp and Scarlet are free to use `deref-or-cancel`, or `deref` with a timeout, as they see fit.

clojure-lsp certainly sees benefit from the client sending it cancellations, but that's largely because the client tends to hammer the server with repeated requests. For example, VS Code is very eager about asking for code actions (gotta keep that 💡  up to date!). It seems to ask for them with nearly every key press. But, as a user keeps typing, it cancels earlier requests, which really does let the server skip some work.

clojure-lsp doesn't initiate many requests, and AFAIK never sends many requests right in a row. I think the only time we do send several in a row is to create a document and then open it. Scarlet does occasionally send several requests in a row, and the client does occasionally have trouble processing all of those and so Scarlet sometimes times out waiting for a response. The server sends a cancellation because it uses `deref-or-cancel`, but I'm not aware of any cases that actually hurts either the client or server.

Anyway, I can see it both ways. 

> Sending a cancel actually adds congestion, as we will block on the output channel to send the cancel notification. 

That's technically true, but I don't see it as being a big problem in practice. The only way we end up blocking on the output channel is if the output stream blocks. But there are several layers of buffers that prevent that from happening—our process, the OS, and the client all have buffers to avoid blocking that stream. If the client has stopped or slowed down reading, and all those buffers fill up, then I think we have bigger problems to fix.

I'll certainly admit that you have better instincts about the potential problems here. This project has lots of concurrency pitfalls. I'm not sure whether that's an architectural problem or it's inherent in the problem space. Anyway, it's worth more effort at some point.

ferdinand-beyer: > Yeah, sounds like a nice addition. I've run out of steam in this PR though. :)

Yeah, sorry for opening up a completely different topic. I'm happy to help, if you want?

Let's put a pin in the `deref-or-cancel` discussion for now.

My only remaining concern is that we now have the chance to return "idiomatic" promises, and with that I mean reject on error, instead of resolving with `result`. I think it is confusing to users who will use the promesa interface of `PendingRequest` but need to implement custom error handling. If we create facts now, we might be hesitant in the future to change this.

I therefore suggest to `p/reject!` the promise on error here:

```suggestion
            (if error
              (p/reject! p (ex-info "Received error response" resp))
              (p/resolve! p result))
```

For backward compatibility, `PendingRequest`'s `deref` and `deref-or-cancel` could use a helper function like this:

```clojure
(defn- resolve-ex-data [p]
  (p/catch p (fn [ex] (or (ex-data ex) (p/rejected ex)))))
```

Calls to `(deref p ,,,)` then change to: `(deref (resolve-ex-data p) ,,,)`.

This way, existing code can work as is, but async code can use the promise API and get what they expect. This is especially nice if users want to integrated `send-request` with oder async code that uses promesa.

We should document that `(deref pending-request)` and `(deref (p/promise pending-request))` report errors differently, but I think this is worth it.

src/lsp4clj/server.clj
ferdinand-beyer: You mean: The code that makes assumptions on how the protocol is implemented is only used by the implementation itself, so it is not designed to work with other implementations?

src/lsp4clj/server.clj
mainej: @ferdinand-beyer sorry for dropping the ball on this. For some reason I thought that we had merged some part of this PR already. Unfortunately, I just merged a Scarlet PR that assumes that this lsp4clj PR has been merged. So now I'm anxious to get this PR merged and released.

I've made the change you suggested. It now seems possible to use the request object as a promesa promise fairly idiomatically. If you'd review the latest commit to ensure it's what you envisioned, I'd appreciate it.

Before you done that, I'm going to start working with @ericdallo to get this merged. @ericdallo would you also do a final review? If you'd like I can remind myself how to merge and do a release, or if you'd be willing to do that, I'd appreciate it.

ericdallo: Sure, I'll take a look again!

ericdallo (approved):
Looks good to me, the code comments seem quite valuable!

mainej:
Thanks @ericdallo. I have a few more small pieces of cleanup but will plan to merge and release this sometime this week.

src/lsp4clj/server.clj
```diff
@@ -96,12 +90,17 @@
...
   `future-cancel` can be called multiple times."
   [id method started server]
-  (map->PendingRequest {:p (promise)
-                        :cancelled? (atom false)
-                        :id id
-                        :method method
-                        :started started
-                        :server server}))
+  ;; Chaining `(-> (p/deferred) (p/catch ...))` seems like it should work, but
+  ;; doesn't.
```
ferdinand-beyer: I think this comment is obsolete, as we identified why we cannot chain: We need to cancel `p`, not the wrapped promise returned by `p/catch`.

ferdinand-beyer (approved):

mainej:
@ericdallo this is merged, but I don't know what `bb` commands to run (perhaps it'd be good to document that), and I'm not sure I have clojars permissions anyway. Will you do a release?

ericdallo:
I'll document that and try to make it work OOTB via github actions @mainej

ericdallo:
@mainej Done, we now have this documented [here](https://github.com/clojure-lsp/lsp4clj#release) and anyone with push permissions can `bb tag x.y.z` and should deploy to clojars using the organization token :)

Also, released as 1.9.2 this PR

mainej:
Amazing! Thanks @ericdallo

PR #42, opened 2023-10-24
Close output channel on error

### LINKED ISSUE #41: Corrupted state when JSON write fails
I've encountered a tricky situation when a request result is not JSON serializable. While this is definitely a programmer's mistake, this unfortunately leads into a situation where the LSP server stops working.

I'm working with VS-Code as a Language Client. After a request, the client reported:

> Connection to server got closed. Server will restart.

Then the client spun up a new server process. The old process was still alive, the JVM up, and my REPL connected. It was somewhat tricky to diagnose what happened.

I tracked this down to the `lsp4clj.io-chan/output-stream->output-chan`, which connects an async channel with a `OutputStream` like this:

```clojure
(let [output (io/output-stream output)
        messages (async/chan 1)]
    (async/thread
      (with-open [writer output] ;; close output when channel closes
        (loop []
          (when-let [msg (async/<!! messages)]
            (write-message writer msg)
            (recur)))))
    messages)
```

When `write-message` fails, because the message is not serializable, an exception is thrown. This code then stops consuming the channel and closes the `OutputStream` through `with-open`. The client then assumes the server died, and will spin up a new one.

However, the `messages` channel will not be closed and will no longer be consumed. That means that the server process will continue to run, and probably hang very soon because of backpressure from the blocked output channel.

ferdinand-beyer:
Fixes #41

This approach closes the output _channel_ when writing a message to the output _stream_ fails for any reason. Expected reasons are:

- I/O exception
- JSON encoding exception

This will prevent the server from blocking forever on a channel that has no more consumers. Instead, messages will be dropped.

We expect that the client will send an exit message and eventually close the connection. This will close the server's input channel and allow it to terminate.

### DIFF
--- a/src/lsp4clj/io_chan.clj
+++ b/src/lsp4clj/io_chan.clj
@@ -128,6 +128,10 @@
       (with-open [writer output] ;; close output when channel closes
         (loop []
           (when-let [msg (async/<!! messages)]
-            (write-message writer msg)
+            (try
+              (write-message writer msg)
+              (catch Throwable e
+                (async/close! messages)
+                (throw e)))
             (recur)))))
     messages))
--- a/test/lsp4clj/io_chan_test.clj
+++ b/test/lsp4clj/io_chan_test.clj
@@ -2,17 +2,33 @@
   (:require
    [clojure.core.async :as async]
    [clojure.string :as string]
-   [clojure.test :refer [deftest is testing]]
+   [clojure.test :refer [deftest is testing use-fixtures]]
    [lsp4clj.io-chan :as io-chan]
    [lsp4clj.test-helper :as h]))
 
 (set! *warn-on-reflection* true)
+
+(defn- silence-uncaught-exceptions [f]
+  (let [handler (Thread/getDefaultUncaughtExceptionHandler)]
+    (Thread/setDefaultUncaughtExceptionHandler
+      (reify Thread$UncaughtExceptionHandler
+        (uncaughtException [_this _thread _e])))
+    (f)
+    (Thread/setDefaultUncaughtExceptionHandler handler)))
+
+(use-fixtures :once silence-uncaught-exceptions)
 
 (defn ^:private message-lines [arr]
   (string/join "\r\n" arr))
 
 (defn mock-input-stream [^String input]
   (.getBytes input "utf-8"))
+
+(defn error-output-stream []
+  (proxy [java.io.OutputStream] []
+    (close [] (throw (java.io.IOException. "close failed")))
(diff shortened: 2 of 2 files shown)


src/lsp4clj/io_chan.clj
```diff
@@ -128,6 +128,10 @@
       (with-open [writer output] ;; close output when channel closes
         (loop []
           (when-let [msg (async/<!! messages)]
-            (write-message writer msg)
+            (try
+              (write-message writer msg)
+              (catch Throwable e
+                (async/close! messages)
+                (throw e)))
```
ericdallo, line 135: I suppose this throw is enough to the developer know this was a JSON parse exception or something right?

ericdallo (approved):
Looks good!

test/lsp4clj/io_chan_test.clj
```diff
@@ -50,6 +66,20 @@
                            "{\"key\":\"äpfel\"}"])
            (.toString output-stream "utf-8")))))
 
+(deftest output-stream-error-should-close-output-channel
+  (testing "when JSON serialisation fails"
+    (let [output-stream (java.io.ByteArrayOutputStream.)
+          output-ch (io-chan/output-stream->output-chan output-stream)]
+      (async/>!! output-ch {:not-serializable output-stream})
```
mainej: suggestion: create a separate non-serializable object.

It appears that the `output-stream` is serving double duty here, both as part of creating the `output-ch`, but also as a non-serializable Object in the message. If that's so, it's clever but also opaque. I'd suggest creating a separate non-serializable thing to use in the message.

mainej (approved):
Seems like a practical solution. One suggestion below, but looks good!

test/lsp4clj/io_chan_test.clj
ferdinand-beyer: Yeah, I was lazy there. Will change that.

src/lsp4clj/io_chan.clj
ferdinand-beyer: I think it depends a bit on the setup. Because this runs within a `core.async/thread`, the exception is probably given to the thread's uncaught exception handler. I think the default one will dump the stack trace to stderr, but many logging frameworks will allow to install a custom handler to log those. Definitely something people should do.

As a library, I figured we should stay un-opinionated here and just throw.

ericdallo: Yeah, agreed

ericdallo:
Thank you!

PR #43, opened 2024-07-19
Complete pending server-initiated request promises on default executor

ferdinand-beyer:
There's a subtle detail on Promesa promises (`CompletableFuture`): Unless an executor is specified, callbacks will be executed on the thread that completed the promise.

For `lsp4clj`, this means that users need to be careful when using `send-request` and coercing the result to a Promesa promise: The code run in handlers such as `p/then` or `p/catch` will end up running in a core.async dispatch thread, as the promise is completed from within a `go-loop`.

Since go blocks are executed in a fixed-size thread pool, one should not block in go blocks. Consequently, users should not block in code that handles the `send-request` promise, unless they explicitly move it to a different executor.

This is easy to get wrong, especially with Promesa's convenience macros like `p/let`, which will coerce the `PendingRequest` to a promise without a dedicated executor. Furthermore, Promesa does not expose `CompletionStage.exceptionallyAsync()`, i.e. it does not support specifying an executor with `p/catch`, therefore it is not trivial to ensure all continuation happens in another thread.

This PR attempts to improve this by completing the promise in a thread from the `:default` executor (`(ForkJoinPool/commonPool)`), instead of the calling thread.

---

Some caveats to this approach:

- There's some overhead even when the promise does not have any continuation handlers (apart from the `CancellationException` we handle ourselves, which, since it blocks on a `send-notification` call, should also run on a suitable executor)
- We don't allow users to customise the executor, but use promesa's default, which is `(ForkJoinPool/commonPool)`. Users might prefer to use a virtual thread executor instead. We could add a `:response-executor` option for this?

---

The change to `.cljfmt.edn` is unrelated, but required for more recent version of cljfmt, as used by Calva, for example. This could be moved to a different PR of course.

### DIFF
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -95,7 +95,7 @@
     ;; client. This cannot be `(-> (p/deferred) (p/catch))` because that returns
     ;; a promise which, when cancelled, does nothing because there's no
     ;; exception handler chained onto it. Instead, we must cancel the
-    ;; `(p/deffered)` promise itself.
+    ;; `(p/deferred)` promise itself.
     (p/catch p CancellationException
       (fn [_]
         (protocols.endpoint/send-notification server "$/cancelRequest" {:id id})))
@@ -351,9 +351,16 @@
         (if-let [{:keys [p started] :as req} (get pending-requests id)]
           (do
             (trace this trace/received-response req resp started now)
-            (if error
-              (p/reject! p (ex-info "Received error response" resp))
-              (p/resolve! p result)))
+            ;; Note that we are called from the server's pipeline, a core.async
+            ;; go-loop, and therefore must not block.  Callbacks of the pending
+            ;; request's promise will be executed in the completing thread,
+            ;; which should not be our thread.  This is very easy for users to
+            ;; miss, therefore we complete the promise on the default executor.
+            (p/thread-call :default
+                           (fn []
+                             (if error
+                               (p/reject! p (ex-info "Received error response" resp))
+                               (p/resolve! p result)))))
           (trace this trace/received-unmatched-response resp now)))
       (catch Throwable e
         (log-error-receiving this e resp))))
--- a/src/lsp4clj/server.clj
+++ b/src/lsp4clj/server.clj
@@ -9,6 +9,7 @@
    [lsp4clj.protocols.endpoint :as protocols.endpoint]
    [lsp4clj.trace :as trace]
    [promesa.core :as p]
+   [promesa.exec :as p.exec]
    [promesa.protocols :as p.protocols])
   (:import
    (java.util.concurrent CancellationException)))
@@ -196,6 +197,7 @@
                        trace-ch
                        tracer*
                        ^java.time.Clock clock
+                       response-executor
                        on-close
                        request-id*
                        pending-sent-requests*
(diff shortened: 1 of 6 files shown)


src/lsp4clj/server.clj
```diff
@@ -351,9 +351,16 @@
         (if-let [{:keys [p started] :as req} (get pending-requests id)]
           (do
             (trace this trace/received-response req resp started now)
-            (if error
-              (p/reject! p (ex-info "Received error response" resp))
-              (p/resolve! p result)))
+            ;; Note that we are called from the server's pipeline, a core.async
+            ;; go-loop, and therefore must not block.  Callbacks of the pending
+            ;; request's promise will be executed in the completing thread,
+            ;; which should not be our thread.  This is very easy for users to
```
mainej: Would the following be an accurate rephrasing? I think this way of putting it requires a little less background on promesa, at the expense of being more verbose.

```suggestion
            ;; go-loop, and therefore must not block. Callbacks of the pending
            ;; request's promise (`p`) will be executed in the completing
            ;; thread, whatever that thread is. Since the callbacks are not
            ;; under our control, they are under our users' control, they could
            ;; block. Therefore, we do not want the completing thread to be our
            ;; thread. This is very easy for users to
```

src/lsp4clj/server.clj
```diff
@@ -351,9 +351,16 @@
...
+            ;; miss, therefore we complete the promise on the default executor.
+            (p/thread-call :default
```
mainej: Is it possible for users to be using the default executor as well? And are we sure that our thread isn't in the default executor?

mainej (approved):
Good catch. See comments for a few questions, but otherwise, looks good to me!

src/lsp4clj/server.clj
ferdinand-beyer: > Is it possible for users to be using the default executor as well?

Not sure if I understand what you mean. Of course they can use the default executor, but there are caveats:

- they would need to remember to pass it to every `p/then` (and friends) on the returned promise
- last time I checked there was no `p/catch` that accepts an executor

It does not matter which thread executes the original `send-request` call, only which thread completes the promise.

> And are we sure that our thread isn't in the default executor?

Yes, as we are called from the `pipeline`, which is a `core.async/go-loop`. Our thread is therefore in core.async's fixed-size `dispatch` thread-pool, that executes all `go` blocks.

ericdallo (approved):
LMK when you 2 think it's ready to merge/release

ferdinand-beyer:
I think this is mergeable. What do you think about letting users override the executor though?

src/lsp4clj/server.clj
mainej: > Our thread is therefore in core.async's fixed-size dispatch thread-pool,

Got it. That makes sense. 

> Not sure if I understand what you mean

I guess my question was whether we're causing problems for our users, by having them to execute their callbacks in the default executor (the common fork-join pool). If they already have a lot of stuff going on in that executor, could it lead to other issues for them? If so, what would we do? Is this materially better or worse for them than what we had before? I lack the familiarity with the underlying tools to answer those questions.

ferdinand-beyer: I think the current behaviour is definitely worse, since right now we run code that can potentially block in a thread pool that is not designed for this. In general, I think `go` blocks are really only meant to move things between channels, everything that requires non-trivial computation or even I/O should be done in threads.

But sure, it would make sense to allow users to specify the executor for responses. I will update this PR to make it configurable.

ericdallo:
> What do you think about letting users override the executor though?

TBH I like it to give a option, if people want to do that, they probably know what they are doing :)

ferdinand-beyer:
@mainej @ericdallo -- I've added a `:response-executor` option, which defaults to `:default` (= `ForkJoinPool/commonPool`). Users can:

- specify `:current-thread` for the old behaviour
- use `:vthread` if they running on a recent JVM
- specify their own executor instance

I've also added tests for that.

ericdallo (approved):
Looks great!

ericdallo:
@ferdinand-beyer could you an entry on CHANGELOG.md please?

ferdinand-beyer:
Done. Also updated the README.

I sneaked in a change to `deps.edn`. I hope that this is OK.

The reason is that with my tooling (Calva), I'd like to have the `test` directory in my classpath, but without _running_ the tests. I normally use two aliases:

- `:test` for configuring the test classpath with `:extra-paths` and `:extra-deps`, enabled during development
- `:test/run` or `:kaocha` to use the test runner of my choice

Did not want to mess with that here, keeping `clojure -M:test` to run the tests, so I added a new `:dev` alias for my needs.

ericdallo:
That's ok add that alias!
Thank you!

ericdallo:
Will wait for https://github.com/clojure-lsp/lsp4clj/pull/44 merge to do a release

ferdinand-beyer:
> Will wait for #44 merge to do a release

Makes sense! Will update that one to resolve the conflict and add the `:response-executor` to the `chan-server` docstring.

PR #47, opened 2025-03-24
Add mcp-clojure-sdk as a known user of lsp4clj

vedang:
I'm the author of [mcp-clojure-sdk](https://github.com/unravel-team/mcp-clojure-sdk).

I've implemented the [MCP](https://modelcontextprotocol.io/introduction) STDIO server using `lsp4clj`, after originally implementing it by hand. `lsp4clj` had already solved a number of issues I stumbled into, and the rewrite on top of lsp4clj was smooth and easy. 

Thank you for an excellent library that made my job so easy.


ericdallo (approved):
Awesome! glad you are using it!

PR #48, opened 2025-05-01
add CI tests

### LINKED ISSUE #4: GithubActions CI for tests
We should have a tests CI for `bb test` and `bb deploy-clojars-*` after merges

frenchy64:
Closes #4 

Proof it works: https://github.com/frenchy64/lsp4clj/pull/1

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,47 @@
+name: Run tests
+
+on:
+  push:
+    branches: [master]
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
+  pull_request:
+    branches: [master]
+
+env:
+  # increment to clear caches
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,47 @@
+name: Run tests
+
+on:
+  push:
+    branches: [master]
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
```
ericdallo, line 7: Don't we want tests to run on every master commit too? not only releases

frenchy64:
Yes, see 2 lines above.

-------- Original Message --------

ericdallo:
Ah I thought that would filter push on master with tag, got it

ericdallo:
Thank you!

